PR #117, opened 2022-11-09
chore: Add to catalog

christopher-cudennec:
Shamelessly copied from https://github.com/SDA-SE/sda-dropwizard-commons/blob/master/catalog-info.yaml

It does not change the system or domain. We should improve these in another step after discussing the best solution.

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -1,0 +1,41 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: sda-spring-boot-commons
+  annotations:
+    sda.se/summary: |
+      The [sda-spring-boot-commons](https://github.com/SDA-SE/sda-spring-boot-commons) library
+      includes several modules to speed up SDA-compliant microservice development. It is
+      based on the [Spring Boot](https://spring.io/projects/spring-boot) framework. It provides
+      several starter modules for additional functionality. Those starter modules configure the
+      framework with default SDA specific settings and add support for an easy and fast integration
+      of standard technologies such as MongoDB, S3 or Kafka.
+    sda.se/federation-scopes: public
+    backstage.io/techdocs-ref: .
+  description: |
+    SDA Spring Boot Commons is a set of libraries to bootstrap services easily that follow
+    the patterns and specifications promoted by the SDA SE.
+spec:
+  type: library
+  lifecycle: production
+  system: sda-spring-boot-commons
+  owner: bedrock
+---
+apiVersion: backstage.io/v1alpha1
+kind: System
+metadata:
+  name: sda-spring-boot-commons
+  description: |
+    SDA Spring Boot Commons is a set of libraries to bootstrap services easily that follow
+    the patterns and specifications promoted by the SDA SE.
+  annotations:
+    sda.se/summary: |
+      The [sda-spring-boot-commons](https://github.com/SDA-SE/sda-spring-boot-commons) library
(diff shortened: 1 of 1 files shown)


eikenilsknopp (dismissed):

brigitteblochsda (commented):
How about adding at least one tag each? e.g. shared:public

catalog-info.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: sda-spring-boot-commons
+  annotations:
```
JoergSiebahn, line 5: You could include [the docs](https://github.com/SDA-SE/sda-spring-boot-commons/blob/master/mkdocs.yaml) as well.

catalog-info.yaml
```diff
@@ -0,0 +1,48 @@
...
+      framework with default SDA specific settings and add support for an easy and fast integration
+      of standard technologies such as MongoDB, S3 or Kafka.
+    sda.se/federation-scopes: public
+  description: |
+    SDA Spring Boot Commons is a set of libraries to bootstrap services easily that follow
+    the patterns and specifications promoted by the SDA SE.
+  links:
+    - url: https://github.com/SDA-SE/sda-spring-boot-commons
+      title: GitHub
+      icon: code
```
JoergSiebahn, line 20: Do we need this link? Imho backstage shows a link to the source anyways.

christopher-cudennec:
> How about adding at least one tag each (for system and library)? e.g. shared:public

@brigitteblochsda : What is this tag used for?
It's no the [documented way](https://sda.dev/developer-guide/development/backstage/share-content-with-the-ecosystem/#ShareContentwiththeEcosystem-MarkSharedContent) to federate the component to the public catalog.

catalog-info.yaml
```diff
@@ -0,0 +1,48 @@
...
+spec:
+  type: library
+  lifecycle: production
+  system: sda-spring-boot-commons
+  owner: bedrock
+---
+apiVersion: backstage.io/v1alpha1
+kind: System
+metadata:
+  name: sda-spring-boot-commons
```
JoergSiebahn, line 30: In the long term we (s|c)ould have a common system for backend libraries (as proposed [here](https://github.com/SDA-SE/sda-kustomize-components/pull/2#discussion_r1011382265)).

Should we prioritize a task to define a structure to avoid moving too many components later?

catalog-info.yaml
christopher-cudennec: Done! 👍

catalog-info.yaml
christopher-cudennec: Agreed. I removed the links.

catalog-info.yaml
christopher-cudennec: It already exists: PLP-471.

brigitteblochsda:
> > How about adding at least one tag each (for system and library)? e.g. shared:public
> 
> @brigitteblochsda : What is this tag used for? It's no the [documented way](https://sda.dev/developer-guide/development/backstage/share-content-with-the-ecosystem/#ShareContentwiththeEcosystem-MarkSharedContent) to federate the component to the public catalog.

@christopher-cudennec: You're right. I just checked other examples and the tag "share:public" seems to be automatically displayed in Backstage in case the is an annotation in the yaml file saying that this repo is public. So never mind. :)

brigitteblochsda (approved):

PR #120, opened 2022-11-18
feat: add ssl support for mongodb

saadmar:
Same as done in dropwizard-commons, this provides the ability to configure the needed certificates at deployment.
One of our services needed this to support documentdb.


sda-commons-cacertificates/src/main/java/org/sdase/commons/cacertificates/CaCertificatesConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+
+  public CaCertificatesConfiguration(
+      @Value("${sda.caCertificates.certificatesDir:/var/trust/certificates}")
+          String defaultTrustedCertificatesDir) {
+    certificateReader = new CertificateReader(defaultTrustedCertificatesDir);
+  }
+
+  @Bean
+  @ConditionalOnMissingBean
+  public SSLContext mongoSSlContext() {
```
JoergSiebahn, line 30: Should we name it without _mongo_ in this context?
I assume, that we will use the same `SSLContext` for clients and if possible for Kafka later.

saadmar: right, renamed

JoergSiebahn (approved):

JoergSiebahn:
Thanks a lot ❤️ @saadmar

PR #137, opened 2022-12-28
fix: enable AutoConfiguration of S3

sdafox:


sda-commons-starter-s3/src/main/java/org/sdase/commons/spring/boot/s3/config/S3Configuration.java
```diff
@@ -47,7 +49,48 @@ public AmazonS3 getAmazonS3Client() {
...
+    this.region = region;
+  }
+
+  public String getBucketName() {
+    return bucketName;
+  }
+
+  public void setBucketName(String bucketName) {
+    this.bucketName = bucketName;
+  }
```
JoergSiebahn, line 95: As these properties are configured by Spring magic, I would not add setters here. Why do we need the getters? Isn't this configuration self contained to provide the S3 Beans?

I see the getters being used in tests. Imho, methods should not be public for tests. You may use package private getters and add `// for test only`.

JoergSiebahn: btw. removing the setters would increase the coverage a lot ;)

JoergSiebahn (approved):

PR #151, opened 2023-01-12
fix: preventing creating tmp dir

elbiocaetano:
BREAKING CHANGE: You need to have a folder called static in the root folder of your container


docs/index.md
```diff
@@ -42,3 +42,7 @@ dependencies {
 
 Artifacts of SDA Spring Boot Commons are available [at Maven Central](https://search.maven.org/search?q=g:org.sdase.commons.spring.boot)
 since release [0.11.2](https://github.com/SDA-SE/sda-spring-boot-commons/releases/tag/0.11.2).
+
+### Static directory
+Since Spring Boot runs in an embedded Tomcat server, it needs a directory to save its files. Therefore, your application need to set a folder called `static` in the root directory of your container.
+In order to do so, create a folder `static` on `src/main/jib`. When the container is generated with `gradlew jibDockerBuild`, this folder will be copied to the container.
```
JoergSiebahn, line 48: NIT: Just a formatting suggestion. Having one sentence per line in markdown makes it easier to review changes in the future and to format the code to a reasonable line length. HTML result will be the same.

```suggestion
### Static directory

Since Spring Boot runs in an embedded Tomcat server, it needs a directory to save its files.
Therefore, your application need to set a folder called `static` in the root directory of your container.
In order to do so, create a folder `static` on `src/main/jib`.
When the container is generated with `gradlew jibDockerBuild`, this folder will be copied to the container.
```

I see that it is hard to describe why we need the static folder. But „it needs a directory to save its files“ is actually not true. But currently I do not find a balance of a short explanation that is not too complex and easy to understand and tells the truth.

Maybe you find better words to express that the folder must be there to support readonly file systems (the main reason for the directory but the opposite of „it needs a directory to save its files“).

JoergSiebahn:
Is it possible to restore [this test](https://github.com/SDA-SE/sda-spring-boot-commons/pull/151/commits/57e9b0c9606a3ea4206fd3159802e066c0b8b8e9#diff-4021a552110a67e36fd0ec433ffdbf19bd2769848312f82d4ab494f5aea74b4bR50)?

I assume, we can't create a `static` dir in the test before Spring starts, but the workdir of the test should be the project root. So we may be able to provide the directory in the codebase. Otherwise it may be an option to adjust the test to accept new directories with docdir in it but no other ones.

sda-commons-web-autoconfigure/src/test/resources/application.properties
```diff
@@ -1,3 +1,5 @@
 management.server.port=0
 wiremock.server.port=0
-spring.zipkin.baseUrl=http://localhost:${wiremock.server.port}/zipkin
\ No newline at end of file
+spring.zipkin.baseUrl=http://localhost:${wiremock.server.port}/zipkin
+
+server.tomcat.basedir=static
```
JoergSiebahn, line 5: Overwriting the property from `defaults.properties` here will imho create a misleading test case.

I tried locally with removing this:

```suggestion
```

and changing the assertion above:

```diff
- assertThat(testApp.getBaseDir()).isEqualTo("static");
+ assertThat(testApp.getBaseDir()).endsWith("/tomcat");
```

the tests are green then.

But anyways, service developers still have to create `/tmp/tomcat` in addition to `/static` in their container, don't they?

docs/index.md
```diff
@@ -42,3 +42,9 @@ dependencies {
 
 Artifacts of SDA Spring Boot Commons are available [at Maven Central](https://search.maven.org/search?q=g:org.sdase.commons.spring.boot)
 since release [0.11.2](https://github.com/SDA-SE/sda-spring-boot-commons/releases/tag/0.11.2).
+
+### Static directory
+Since Spring Boot runs in an embedded Tomcat server, it needs a tmp directory to support the container run in a readonly file system.
+Therefore, your application need to set a folder called `static` in the root directory of your container.
+In order to do so, create a folder `static` on `src/main/jib`.
+When the container is generated with `gradlew jibDockerBuild`, this folder will be copied to the container.
```
JoergSiebahn, line 50: I like the explanation!

But don't service developers still have to create `/tmp/tomcat` in addition to `/static` in their container?

sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/security/SystemDependentBaseDirTest.java
```diff
@@ -0,0 +1,107 @@
...
+  private static Path createStaticDirectory() {
+    try {
+      Path root = Paths.get(".").normalize().toAbsolutePath();
+      Path filePath = Paths.get(root.toString(), "static");
+
+      deleteStaticDirectory(filePath);
+
+      return Files.createDirectory(filePath);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
JoergSiebahn, line 78: Minor optimization to convert `IOException` into an unchecked exception with more meaning:

```suggestion
      throw new UncheckedIOException(e);
```

docs/index.md
```diff
@@ -42,3 +42,14 @@ dependencies {
...
+### Static directories
+Since Spring Boot runs in an embedded Tomcat server, it needs some tmp directories to support the container run in a readonly file system.
+Therefore, your application need to set a folder called `static` and a folder structure `tmp/tomcat` in the root directory of your container.
+In order to do so, create a folder `static` and `tmp/tomcat` on `src/main/jib`.
+When the container's image is generated with `gradlew jibDockerBuild`, these folders will be copied to the container.
+In case you need the tmp folder to be writable, you can mount a volume in your container and set the environment variable pointing to your folder:
+
+```
+SERVER_TOMCAT_BASEDIR=/tmp/tomcat
+```
```
JoergSiebahn, line 55: Isn't this covered by the [predefined property](https://github.com/SDA-SE/sda-spring-boot-commons/pull/151/files#diff-948deeb2743800b0019af9f1f614f42abc071037669b91c7b23840b27eddc303R5)?

If so:
```suggestion
In case you need the tmp folder to be writable, you can mount a writable volume to `/tmp` in your container.
```

sda-commons-web-autoconfigure/src/main/resources/META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports
```diff
@@ -9,4 +9,4 @@ org.sdase.commons.spring.boot.web.client.SdaOidcClientConfiguration
 org.sdase.commons.spring.boot.web.docs.SdaOpenApiCustomizerConfiguration
 org.sdase.commons.spring.boot.web.jackson.SdaObjectMapperConfiguration
 org.sdase.commons.spring.boot.web.monitoring.SdaMonitoringConfiguration
-org.sdase.commons.spring.boot.web.monitoring.SdaTracingConfiguration
+org.sdase.commons.spring.boot.web.monitoring.SdaTracingConfiguration
```
JoergSiebahn, line 12: Could you undo this change? It may cause some trouble with other open PRs.

docs/index.md
elbiocaetano: I will adjust to say that they can change to another location, if they prefer.

JoergSiebahn (approved):

JoergSiebahn (approved):

PR #163, opened 2023-01-18
Document security hardening and enable security for sda platform

JoergSiebahn:
- [x] wait for #162 to be merged

Note: Security is already available for auto configured services. Users of `@EnableSdaPlatform` get the security features as well now.

PR #165, opened 2023-01-20
Max post size

JoergSiebahn:
- [x] wait for #163 to be merged

### DIFF
--- a/sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/security/LimitRequestBodySizeFilter.java
+++ b/sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/security/LimitRequestBodySizeFilter.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2022- SDA SE Open Industry Solutions (https://www.sda.se)
+ *
+ * Use of this source code is governed by an MIT-style
+ * license that can be found in the LICENSE file or at
+ * https://opensource.org/licenses/MIT.
+ */
+package org.sdase.commons.spring.boot.web.security;
+
+import java.io.IOException;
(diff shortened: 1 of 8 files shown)


sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/security/LimitRequestBodySizeFilter.java
```diff
@@ -0,0 +1,57 @@
...
+
+  @Override
+  protected void doFilterInternal(
+      HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
+      throws ServletException, IOException {
+    if (request.getContentLengthLong() > requestBodyMaxSize.toBytes()) {
+      response.sendError(HttpStatus.PAYLOAD_TOO_LARGE.value()); // 413
+    } else if (isChunkedEncoding(request.getHeaders(HttpHeaders.TRANSFER_ENCODING))) {
+      response.sendError(HttpStatus.LENGTH_REQUIRED.value()); // 411
+    } else {
```
christopher-cudennec, line 42: Wouldn't it be easier to add a test for a missing CONTENT_LENGTH header? Or we could add it as an additional condition.

JoergSiebahn: That would break all GET, HEAD and OPTIONS requests and most (close to all) DELETE requests. But the RFC does not strictly forbid payload for these requests. So it would lead to a lot more conditions to separate the methods. It felt just less complex and less prone to errors to verify the chunked encoding in general.

christopher-cudennec (approved):

PR #185, opened 2023-02-15
fix: use default trust manager if no custom certificates are mounted

saadmar:
Spring boot does not allow beans to be null, this will use only the default trust store even if no certificate is mounted in the configured path.
The current behaviour is that the application will fail to start if no certificates are provided

- [x] Test in a [service](https://github.com/SDA-SE/email-transformation-orchestrator/pull/29/commits/6542bfa3d72758da3cd4f962f374b82ef40adb9f)

### DIFF
--- a/sda-commons-cacertificates/src/test/java/org/sdase/commons/cacertificates/CaCertificatesConfigurationTest.java
+++ b/sda-commons-cacertificates/src/test/java/org/sdase/commons/cacertificates/CaCertificatesConfigurationTest.java
@@ -14,9 +14,7 @@ import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
 
-@SpringBootTest(
-    classes = TestApp.class,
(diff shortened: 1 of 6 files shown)


sda-commons-cacertificates/src/test/java/org/sdase/commons/cacertificates/CaCertificatesConfigurationTest.java
```diff
@@ -14,9 +14,7 @@
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
 
-@SpringBootTest(
-    classes = TestApp.class,
-    properties = {"sda.caCertificates.certificatesDir=src/test/resources"})
+@SpringBootTest(classes = TestApp.class)
```
JoergSiebahn, line 17: Could we duplicate this test to verify bean creation with and without config?

saadmar: done 👍

JoergSiebahn (approved):

PR #187, opened 2023-02-15
fix: health check for open policy agent

elbiocaetano:


docs/web/index.md
```diff
@@ -528,6 +529,7 @@ management.endpoint.health.probes.enabled=true
 management.endpoint.health.group.readiness.include=readinessState, mongo
 management.endpoint.health.show-details=always
 management.endpoint.health.show-components=always
+management.health.openPolicyAgent.enabled=true
```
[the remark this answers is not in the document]
elbiocaetano: No, this is for the bean itself, not for the endpoint. This is used by the annotation `@ConditionalOnEnabledHealthIndicator`

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/monitoring/SdaMonitoringConfiguration.java
```diff
@@ -7,9 +7,21 @@
...
 @PropertySource("classpath:/org/sdase/commons/spring/boot/web/monitoring/monitoring.properties")
-public class SdaMonitoringConfiguration {}
+public class SdaMonitoringConfiguration {
+
+  @Bean
+  @ConditionalOnProperty(name = "opa.disable", havingValue = "false", matchIfMissing = true)
+  @ConditionalOnEnabledHealthIndicator("openPolicyAgent")
+  public OpenPolicyAgentHealthIndicator openPolicyAgentHealthContributor() {
+    return new OpenPolicyAgentHealthIndicator();
+  }
```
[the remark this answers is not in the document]
elbiocaetano: But why would you want to create the bean if opa is disabled? Would not it consume more memory with a bean that would be never used? With the ` @ConditionalOnEnabledHealthIndicator` annotation you can even disable this bean even with the opa.disabled enabled, in case you do not want to perform  the health check

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/auth/opa/OpaAccessDecisionVoter.java
```diff
@@ -159,12 +159,12 @@ private int handleOpaDisabled(HttpServletRequest httpRequest) {
     return ACCESS_GRANTED;
   }
 
-  private String createOpaRequestUri(String opaBaseUrl, String policyPackage) {
+  public static String createOpaRequestUri(String opaBaseUrl, String policyPackage) {
```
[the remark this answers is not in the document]
elbiocaetano: I like the idea. Done ✔️

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/monitoring/SdaMonitoringConfiguration.java
```diff
@@ -7,9 +7,28 @@
...
 @AutoConfiguration
 @PropertySource("classpath:/org/sdase/commons/spring/boot/web/monitoring/monitoring.properties")
-public class SdaMonitoringConfiguration {}
+public class SdaMonitoringConfiguration {
+
+  @Value("${opa.base.url:http://localhost:8181}")
+  private String opaBaseUrl;
+
+  @Value("${opa.policy.package:}")
+  private String policyPackage;
```
JoergSiebahn, line 26: To me it seems like this properties should not be in this place.

Is it possible to create the `OpenPolicyAgentHealthIndicator` as bean with the `OpaAccessDecisionVoter` as dependency and add a package private getter for `opaRequestUrl`? Then we don't even need to add it to this configuration.

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/monitoring/SdaMonitoringConfiguration.java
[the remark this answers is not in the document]
JoergSiebahn: I wonder why we should make it configurable to have the health check. As long as OPA is enabled, the health check should be active. Also [the docs](https://github.com/SDA-SE/sda-spring-boot-commons/pull/187/files#diff-2ead3b4ae5e0196eaed448f28100ca5495e172e2b954707c5c52ce2006f09f84R489) could be simplified to only _if OPA is enabled_. Then we don't need to bloat the docs with the defaults.

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/monitoring/SdaMonitoringConfiguration.java
elbiocaetano: So if the AccessDecisionVoter is a dependency and it has the requestUrl and the opaRequestTemplate as well, I created a method there to call the get endpoint as package private, so I can use in my class as well

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/monitoring/SdaMonitoringConfiguration.java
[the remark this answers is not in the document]
elbiocaetano: I think we should provide configuration options to the user, so they can choose where having the health check or not. but thinking in this specific case it does not make sense to disable the health check, if the opa is enabled, because the pod will be marked as ready, but the requests will fail. So I removed this condition and left just the opa.disabled property check

docs/web/index.md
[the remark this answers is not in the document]
elbiocaetano: It was removed based on comment https://github.com/SDA-SE/sda-spring-boot-commons/pull/187#discussion_r1108406910

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/auth/opa/OpaAccessDecisionVoter.java
```diff
@@ -152,19 +152,23 @@ private OpaResponse authorizeWithOpa(HttpServletRequest httpRequest) {
...
+
   private int handleOpaDisabled(HttpServletRequest httpRequest) {
     if (httpRequest.getUserPrincipal() == null) {
       LOG.warn("OPA is disabled. Access is granted for anonymous user without constraints.");
     }
     return ACCESS_GRANTED;
   }
 
-  private String createOpaRequestUri(String opaBaseUrl, String policyPackage) {
+  static String createOpaRequestUri(String opaBaseUrl, String policyPackage) {
```
JoergSiebahn, line 166: Can be reverted

```suggestion
  private String createOpaRequestUri(String opaBaseUrl, String policyPackage) {
```

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/auth/opa/OpaAccessDecisionVoter.java
```diff
@@ -152,19 +152,23 @@ private OpaResponse authorizeWithOpa(HttpServletRequest httpRequest) {
...
     return String.format(
         "%s/v1/data/%s", opaBaseUrl.trim(), createPolicyPathFromPackage(policyPackage));
   }
 
-  private String createPolicyPathFromPackage(String policyPackage) {
+  private static String createPolicyPathFromPackage(String policyPackage) {
```
JoergSiebahn, line 171: Can be reverted

```suggestion
  private String createPolicyPathFromPackage(String policyPackage) {
```

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/auth/opa/OpaAccessDecisionVoter.java
```diff
@@ -40,9 +40,9 @@ public class OpaAccessDecisionVoter implements AccessDecisionVoter<FilterInvocat
   private static final Logger LOG = LoggerFactory.getLogger(OpaAccessDecisionVoter.class);
 
   private final boolean disableOpa;
-  private final String opaRequestUrl;
+  final String opaRequestUrl;
```
JoergSiebahn, line 43: Can be reverted?

```suggestion
  private final String opaRequestUrl;
```

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/auth/opa/OpaAccessDecisionVoter.java
```diff
@@ -40,9 +40,9 @@ public class OpaAccessDecisionVoter implements AccessDecisionVoter<FilterInvocat
...
   private final OpaRequestBuilder opaRequestBuilder;
-  private final RestTemplate opaRestTemplate;
+  final RestTemplate opaRestTemplate;
```
JoergSiebahn, line 45: Can be reverted?

```suggestion
  private final RestTemplate opaRestTemplate;
```

docs/web/index.md
```diff
@@ -486,6 +486,13 @@ The readiness group contains the following indicators:
 
 *   `ReadinessStateHealthIndicator`
 *   `MongoHealthIndicator`, if auto-configured.
+*   `OpenPolicyAgentHealthIndicator` if the property `opa.disable=false` or not set.
+* 
+### Default properties
+
+```properties
+opa.disable=false
+```
```
JoergSiebahn, line 495: Imho the non-technical part of the condition is easier to understand

```suggestion
*   `OpenPolicyAgentHealthIndicator` if [OPA](#opa) is enabled for authentication

```

JoergSiebahn (approved):

JoergSiebahn (approved):

PR #192, opened 2023-02-22
feat: Provide handling of a MetadataContext to track metadata across services for synchronous, asynchronous and interrupteds business processes

eikenilsknopp:


sda-commons-dependency-check/src/test/java/org/sdase/commons/spring/boot/dependency/check/DuplicateClassesTest.java
```diff
@@ -43,6 +44,7 @@ class DuplicateClassesTest {
    * href="https://stackoverflow.com/a/52639079">Stackoverflow</a>
    */
   @Test
+  @Disabled
```
eikenilsknopp, line 47: Excluded for now

sda-commons-metadata-context/build.gradle
```diff
@@ -0,0 +1,16 @@
+dependencies {
+  api 'org.springframework.cloud:spring-cloud-starter-openfeign'
+  api 'org.springframework.boot:spring-boot-starter-web'
```
christopher-cudennec, line 3: That means that all modules that depend on "sda-commons-metadata-context" will still get all the web dependencies. Have we achieved the goal of the refactoring? Would probably result in more modules if we want to separate "metadata-context-core" + "metadata-context-web". What do we prefer?

elbiocaetano: Resolved on https://github.com/SDA-SE/sda-spring-boot-commons/pull/211

JoergSiebahn (approved):

PR #193, opened 2023-02-23
Dependency check

JoergSiebahn:


sda-commons-web-testing/build.gradle
```diff
@@ -1,6 +1,14 @@
 dependencies {
   api 'org.springframework.boot:spring-boot-starter-test'
-  api 'org.springframework.cloud:spring-cloud-contract-wiremock'
+  api 'org.springframework.cloud:spring-cloud-contract-wiremock', {
```
christopher-cudennec, line 3: How does that fix version conflicts? I expected to see an exclude/include for "commons-upload" here.

JoergSiebahn: `wiremock-jre8-standalone` repackages classes from other libraries that we already have as dependency.

christopher-cudennec: Ah, that's the difference. We replace `wiremock-jre8-standalone` by `wiremock-jre8`. 👍

christopher-cudennec (approved):

PR #198, opened 2023-02-27
feat: implementing metadata for rest calls

elbiocaetano:


sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/metadata/ContainerRequestContextHolder.java
```diff
@@ -0,0 +1,110 @@
...
+      new ThreadLocal<>();
+
+  /**
+   * @return The current {@link ContainerRequestContext} if the current {@link Thread} is in a
+   *     request context. An empty {@code Optional} if the current thread is not in a request
+   *     context.
+   */
+  static Optional<ContainerRequestContext> currentRequestContext() {
+    return Optional.ofNullable(REQUEST_CONTEXT_HOLDER.get());
+  }
```
christopher-cudennec, line 41: Can't we use Spring's [RequestContextHolder](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/context/request/RequestContextHolder.html)?

JoergSiebahn: Imho this is not needed in the Spring context. The important part of transferring the metadata context should be included in the [ContextCopyTaskDecorator](https://github.com/SDA-SE/sda-spring-boot-commons/blob/master/sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/async/ContextCopyTaskDecorator.java). As @christopher-cudennec said, Spring's RequestContextHolder should be already doing what's implemented here.

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/metadata/MetadataRequestInterceptor.java
```diff
@@ -0,0 +1,64 @@
...
+  public MetadataRequestInterceptor(Set<String> metadataFields) {
+    this.metadataFields = metadataFields;
+  }
+
+  @Override
+  public void afterCompletion(
+      HttpServletRequest request, HttpServletResponse response, Object handler, Exception ex)
+      throws Exception {
+    MetadataContextHolder.clear();
+    HandlerInterceptor.super.afterCompletion(request, response, handler, ex);
```
JoergSiebahn, line 33: As the method is just an empty default for convenience, we don't need to call super. Or am I missing something?

```suggestion
```

sda-commons-web-autoconfigure/build.gradle
```diff
@@ -35,4 +35,8 @@ dependencies {
   testImplementation 'org.junit-pioneer:junit-pioneer:2.0.0', {
     exclude group: 'org.junit'
   }
+
+  test {
+    environment "METADATA_FIELDS", "tenant-id,processes"
+  }
```
JoergSiebahn, line 41: I would turn it into a configuration property and make MetadataContext a bean. Then you can define the override in the test. As this PR is not about „springification“ that's maybe not right place to do it here. But could you leave a TODO so that we can use the standard way of Spring Boot configuration later?

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/metadata/MetadataClientInterceptor.java
```diff
@@ -7,40 +7,36 @@
...
-import javax.ws.rs.client.ClientRequestContext;
-import javax.ws.rs.client.ClientRequestFilter;
 import org.apache.commons.lang3.StringUtils;
 
-/**
- * A {@link ClientRequestFilter} to submit the {@link MetadataContext} to other services that are
- * called synchronously.
- */
-public class MetadataContextClientRequestFilter implements ClientRequestFilter {
+public class MetadataClientInterceptor implements RequestInterceptor {
```
christopher-cudennec, line 16: Can we use `MetadataContext` as prefix in all cases?
I.e. from `MetadataClientInterceptor` to `MetadataContextClientInterceptor`

christopher-cudennec: We should also keep the documentation.

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/metadata/MetadataRequestInterceptor.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Enumeration;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Stream;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.web.servlet.HandlerInterceptor;
+
+public class MetadataRequestInterceptor implements HandlerInterceptor {
```
christopher-cudennec, line 20: Also use `MetadataContext` as prefix here?
The class could also use some short documentation.

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/metadata/MetadataContextRequestInterceptor.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+    return metadataContext;
+  }
+
+  private List<String> normalize(Enumeration<String> original) {
+    if (original == null) {
+      return List.of();
+    }
+    return Collections.list(original).stream()
+        .map(s -> s.split(","))
```
JoergSiebahn, line 80: Maybe I'm a bit too careful here, but as we can't be 100% sure, what's in the headers:

```suggestion
    return Collections.list(original).stream()
        .filter(StringUtils::isNotBlank)
        .map(s -> s.split(","))
```

sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/metadata/MetadataOtherServiceAuthenticatedClient.java
```diff
@@ -0,0 +1,26 @@
...
+import org.sdase.commons.spring.boot.web.client.AuthenticationPassThroughClientConfiguration;
+import org.springframework.cloud.openfeign.FeignClient;
+import org.springframework.http.HttpHeaders;
+import org.springframework.web.bind.annotation.GetMapping;
+
+@FeignClient(
+    value = "metadataOtherAuthenticated",
+    url = "${metadata.otherAuthenticated.baseUrl}",
+    configuration = {AuthenticationPassThroughClientConfiguration.class})
+public interface MetadataOtherServiceAuthenticatedClient {
```
JoergSiebahn, line 19: Is the metadata context added unconditionally to all clients? Imho that should be only used for clients that are calling with the platform. Like the authentication pass through is configured here with `AuthenticationPassThroughClientConfiguration`. That should also be documented somewhere, like in `EnableSdaClients` or [the docs](https://github.com/SDA-SE/sda-spring-boot-commons/blob/b5aa5acb20a9e454c3ebcb718846f3e2c97bb825/docs/web/index.md#http-client) (which are actually missing the trace token config).

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/metadata/MetadataContextRequestInterceptor.java
christopher-cudennec: +1 - are null values possible? This should be covered by `isNotBlank`.

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/metadata/MetadataContextTokenConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.web.servlet.HandlerInterceptor;
+import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
+
+@AutoConfiguration
+public class MetadataContextTokenConfiguration implements WebMvcConfigurer {
+
+  @Bean
+  public HandlerInterceptor metadataRequestInterceptor() {
```
christopher-cudennec, line 21: Also use `metadataContext...` as the bean names (also see below)?

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/metadata/MetadataContextTokenConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+
+import feign.RequestInterceptor;
+import org.springframework.boot.autoconfigure.AutoConfiguration;
+import org.springframework.context.annotation.Bean;
+import org.springframework.web.servlet.HandlerInterceptor;
+import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
+
+@AutoConfiguration
+public class MetadataContextTokenConfiguration implements WebMvcConfigurer {
```
christopher-cudennec, line 18: I don't think we used `token` anywhere. Do you have a good reason for it?
Otherwise I'd just go with `MetadataContextConfiguration`.

sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/metadata/ApiClientConfigurationTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+    mockMvc
+        .perform(
+            MockMvcRequestBuilders.request(HttpMethod.GET, "/metadataAuthProxy")
+                .headers(httpHeaders)
+                .accept(APPLICATION_JSON))
+        .andExpect(MockMvcResultMatchers.status().isOk());
+    verify(metadataOtherServiceAuthenticatedClient).getSomething(httpHeaders);
+
+    MetadataContext.createContext(new DetachedMetadataContext());
```
christopher-cudennec, line 72: I'm not 100% familiar with our API. Don't we have a `clear` or `reset` method?
Should we move this line to a `beforeEach` method?

sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/metadata/ApiClientConfigurationTest.java
```diff
@@ -0,0 +1,98 @@
...
+        .thenReturn(Map.of("hello", "world"));
+
+    mockMvc
+        .perform(
+            MockMvcRequestBuilders.request(HttpMethod.GET, "/metadataProxy")
+                .headers(httpHeaders)
+                .accept(APPLICATION_JSON))
+        .andExpect(MockMvcResultMatchers.status().isOk());
+
+    verify(metadataOtherServiceClient).getSomething(httpHeaders);
```
christopher-cudennec, line 94: Don't we expect blank headers to be sent to other/external services? `httpHeaders` here is the same as in the test above.

sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/metadata/ApiClientConfigurationTest.java
JoergSiebahn: We have it [for internal use](https://github.com/SDA-SE/sda-spring-boot-commons/blob/PLP-698-implement-for-rest-calls/sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/metadata/MetadataContextHolder.java#L26). That would be accessible in this test. I would also prefer `@AfterEach` and/or `@BeforeEach`, but at least a `finally`.

sda-commons-web-autoconfigure/build.gradle
JoergSiebahn: As you use the `MetadataContext.metadataFields()` now, you could also add `@SetSystemProperties` to the test classes (at the very top). Example: https://github.com/SDA-SE/sda-spring-boot-commons/pull/200/files#diff-a059385d2de54b2c378821272fc0622d23ff8a57485c6af731579bef425c92ddR39

JoergSiebahn (approved):

PR #201, opened 2023-03-01
fix(jackson): write invalid Json when errors happen while serialization

JoergSiebahn:
These errors are the only way to identify a serialisation error on consumer side when e.g. a Json is written directly to the response output stream. Streaming data directly from a source (e.g. a MongoCursor) into the response as chunked encoding is valuable to minimize memory resources but may lead to not identifiable errors (e.g. while mapping from entity model to API model) when the ObjectMapper creates a valid Json after the error when the generator is closed.

### DIFF
--- a/sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/jackson/ObjectMapperTest.java
+++ b/sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/jackson/ObjectMapperTest.java
@@ -7,11 +7,19 @@
  */
 package org.sdase.commons.spring.boot.web.jackson;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.ByteArrayOutputStream;
+import java.io.Closeable;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
 import org.junit.jupiter.api.Test;
 import org.sdase.commons.spring.boot.web.jackson.test.Filter;
 import org.sdase.commons.spring.boot.web.jackson.test.JacksonTestApp;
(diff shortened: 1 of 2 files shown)


sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/jackson/ObjectMapperTest.java
```diff
@@ -154,4 +161,47 @@ void shouldNotFailForUnknownSubtype() throws Exception {
...
+    }
+    var expected = new StringBuilder();
+    for (int i = 1; i < 1_001; i++) {
+      expected.append("\"").append(i).append("\",");
+    }
+    expected.deleteCharAt(expected.length() - 1);
+    assertThat(output.toString(UTF_8)).isEqualToIgnoringWhitespace("{\"items\":[" + expected);
+  }
+
+  static class CloseableIterator implements Iterator<String>, Closeable {
```
christopher-cudennec, line 184: Better name `FailingIterator` to make sure what the class should do?

sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/jackson/ObjectMapperTest.java
```diff
@@ -154,4 +161,47 @@ void shouldNotFailForUnknownSubtype() throws Exception {
...
+      generator.writeObject(Map.of("items", iterator));
+    } catch (Exception ignored) {
+      // nothing to do
+    }
+    var expected = new StringBuilder();
+    for (int i = 1; i < 1_001; i++) {
+      expected.append("\"").append(i).append("\",");
+    }
+    expected.deleteCharAt(expected.length() - 1);
+    assertThat(output.toString(UTF_8)).isEqualToIgnoringWhitespace("{\"items\":[" + expected);
```
christopher-cudennec, line 181: Do we want to ensure that `output` is not a valid object?

JoergSiebahn: added an assert

sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/jackson/ObjectMapperTest.java
JoergSiebahn: changed to `FailingClosableIterator`

christopher-cudennec (approved):

PR #204, opened 2023-03-01
fix(client): create trace-token for outgoing requests when not in a request context

### LINKED ISSUE #203: Request interceptors error: No thread-bound request found
Hello,
I'm currently facing one issue which seems to be related to the RequestInterceptors used for tracing and logging of HTTP Requests.
Example:
[        RequestContextHolder.currentRequestAttributes()](https://github.com/SDA-SE/sda-spring-boot-commons/blob/3ab385c398340db74b0ad3394215ad4a29d1690c/sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/tracing/TraceTokenClientInterceptor.java#L28)

Context: i'm using a thread pool to fetch & execute some tasks which call other REST services to manipulate data 

The error I have is the one fired from Spring RequestContextHolder.currentRequestAttributes() because the thread which execute the tasks does not have any request atttributes

image

Do you have any idea how this issue can be fixed ?

Kind Regards,
Dragos

JoergSiebahn:
fixes #203

### DIFF
--- a/sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/tracing/TraceTokenClientInterceptor.java
+++ b/sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/tracing/TraceTokenClientInterceptor.java
@@ -12,10 +12,14 @@ import static org.sdase.commons.spring.boot.web.tracing.TraceTokenRequestInterce
 import feign.RequestInterceptor;
 import feign.RequestTemplate;
 import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.web.context.request.RequestAttributes;
 import org.springframework.web.context.request.RequestContextHolder;
 
 public class TraceTokenClientInterceptor implements RequestInterceptor {
+
+  private static final Logger LOG = LoggerFactory.getLogger(TraceTokenClientInterceptor.class);
 
   @Override
   public void apply(RequestTemplate template) {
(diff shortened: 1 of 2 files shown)


sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/tracing/TraceTokenClientInterceptor.java
```diff
@@ -12,26 +12,33 @@
...
-            .getAttribute(TRACE_TOKEN_HEADER_NAME, RequestAttributes.SCOPE_REQUEST);
-
-    if (attribute instanceof String strAttribute) {
-      RequestContextHolder.currentRequestAttributes()
-          .setAttribute(TRACE_TOKEN_HEADER_NAME, strAttribute, RequestAttributes.SCOPE_REQUEST);
-      return strAttribute;
+            .setAttribute(TRACE_TOKEN_HEADER_NAME, strAttribute, RequestAttributes.SCOPE_REQUEST);
+        return strAttribute;
+      }
+    } catch (IllegalStateException e) {
```
christopher-cudennec, line 40: It's unfortunate that there is no clean way to find out if we're in a request context instead of catching the exception.

As alternative we could invoke `RequestContextHolder#getRequestAttributes` which would return `null` if the attributes are missing. Still it's an indirect way to find out and the implementation is slighty different because it does not fall back to `FacesRequestAttributesFactory` in case of JSF.

JoergSiebahn: Just from the wording, without knowing the actual implementation of the foreign library, _current_ sounds much more reasonable than _get_. What's so bad about catching a documented exception?

christopher-cudennec: Just that it should be an "exception" and not a regular use case :D 
But I'm fine with your approach since there is no better alternative like `hasRequestContext`.

christopher-cudennec (approved):

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/tracing/TraceTokenClientInterceptor.java
JoergSiebahn: Well, first checking and then accessing can lead to edge cases. Maybe not for something that is bound to the current thread like here. But in many other cases, checking first and then accessing is not a good idea. Examples: Accessing existing or non-existing files after checking _exists_, modifying data in a non-transactional database after checking if it matches a specific criteria. There could always be another process modifying it at the same time.

PR #205, opened 2023-03-02
feat(kafka): allow `concurrency` configuration for `kafka consumer`

vladimirsvicevicsrb:


christopher-cudennec (commented):
Looks pretty straightforward to me. Are we able to add a test?

vladimirsvicevicsrb:
> Looks pretty straightforward to me. Are we able to add a test?

I didn't come up with an idea on how to test that locally, as it looks like I am testing spring internals. Any idea would be appreciated.
At least, I ran existing tests found in `sda-spring-boot-commons` and they pass.

I prepared two PRs:
* https://github.com/SDA-SE/medical-claim-ods/pull/63
* https://github.com/SDA-SE/gitops-health-stack/pull/269
in order to test one ODS with `PR snapshot`

sda-commons-starter-kafka/src/main/java/org/sdase/commons/spring/boot/kafka/config/KafkaConsumerConfig.java
```diff
@@ -13,7 +13,7 @@
 
 @ConstructorBinding
 @ConfigurationProperties(prefix = "sda.kafka.consumer")
-public record KafkaConsumerConfig(@NotNull RetryConfig retry) {
+public record KafkaConsumerConfig(@NotNull RetryConfig retry, @NotNull Integer concurrency) {
```
christopher-cudennec, line 16: Does it make more sense to use `int` if it's never `null`?

christopher-cudennec (commented):
What will happen if applications upgrade? Will they get the default value of `1` from the `consumer.properties`?

sda-commons-starter-kafka/src/main/java/org/sdase/commons/spring/boot/kafka/config/KafkaConsumerConfig.java
vladimirsvicevicsrb: I guess the same as for the other `RetryConfig`, as an example: `@NotNull Integer maxRetries`
I was thinking that config [here](https://github.com/SDA-SE/sda-spring-boot-commons/pull/205/files/67bf4c79a3e693a8dcfcd62b52520162c9c741cf#diff-9ea7b4e4591d6c14e08b6031d5b6bbe222e09b9cbc56c9b81a0732bdc005770e) is the default one - no?

christopher-cudennec: maybe @JoergSiebahn can help out here 😃

elbiocaetano: Yes, it will read from the application.properties, because of the annotation `@ConfigurationProperties(prefix = "sda.kafka.consumer")`
Also, the default concurrency for kafka concurrency is 1, so the value is correctly in my opinion.

elbiocaetano:
> > Looks pretty straightforward to me. Are we able to add a test?
> 
> I didn't come up with an idea on how to test that locally, as it looks like I am testing spring internals. Any idea would be appreciated. At least, I ran existing tests found in `sda-spring-boot-commons` and they pass.
> 
> I prepared two PRs:
> 
> * [test: test kafka consumer concurrency configuration medical-claim-ods#63](https://github.com/SDA-SE/medical-claim-ods/pull/63)
> * [test: test kafka consumer concurrency configuration gitops-health-stack#269](https://github.com/SDA-SE/gitops-health-stack/pull/269)
>   in order to test one ODS with `PR snapshot`

You could create a test to see if the properties are being set, like in the test `org.sdase.commons.spring.boot.s3.config.S3ConfigurationTest`.
I agree that we don’t need to test the library, so I would just add a test to check whether the config is being correctly set or not

sda-commons-starter-kafka/src/main/java/org/sdase/commons/spring/boot/kafka/config/KafkaConsumerConfig.java
vladimirsvicevicsrb: With the current approach you are consuming data from only one thread regardless of number of partitions you have in the topic.

This concurrency config is a way to say to spring: please create 3 threads in this consumer and distribute them among available partitions so that the performance is increased.
So:
```Topic: test, partitions: 0,1,2
* java-thread-1 consumers from partition 0
* java-thread-2 consumers from partition 1
* java-thread-3 consumers from partition 2
```
In case there are 6 partitions then spring will orchestrate them among available thread defined via concurrency flag, smth like:
```Topic: test, partitions: 0,1,2,3,4,5
* java-thread-1 consumers from partition 0,3
* java-thread-2 consumers from partition 1,4
* java-thread-3 consumers from partition 2,5
```

I am thinking about how to accomplish this by spinning up multiple replicas. I guess we would need to use same `consumer-group-id` on all replicas and probably to instruct them to consumer from different partitions?

Also, I am thinking about whether it is easier, that for the beginning - until we don't know the resource requirements, we start by having 1 replica set, and them horizontally scale if it is really needed. Maybe only `kafka` part needs more power and `read part - api` is completely fine. I am also not sure whether it is more performant to spin up 3 jvms instead of 3 threads.

My assumption is based on already provided functionality withing spring-kafka ecosystem that's is not configurable withing sda spring-boot service.

vladimirsvicevicsrb:
> > > Looks pretty straightforward to me. Are we able to add a test?
> > 
> > 
> > I didn't come up with an idea on how to test that locally, as it looks like I am testing spring internals. Any idea would be appreciated. At least, I ran existing tests found in `sda-spring-boot-commons` and they pass.
> > I prepared two PRs:
> > 
> > * [test: test kafka consumer concurrency configuration medical-claim-ods#63](https://github.com/SDA-SE/medical-claim-ods/pull/63)
> > * [test: test kafka consumer concurrency configuration gitops-health-stack#269](https://github.com/SDA-SE/gitops-health-stack/pull/269)
> >   in order to test one ODS with `PR snapshot`
> 
> You could create a test to see if the properties are being set, like in the test `org.sdase.commons.spring.boot.s3.config.S3ConfigurationTest`. I agree that we don’t need to test the library, so I would just add a test to check whether the config is being correctly set or not

Thanks for the hint. I will have a look into it.

sda-commons-starter-kafka/src/main/java/org/sdase/commons/spring/boot/kafka/config/KafkaConsumerConfig.java
JoergSiebahn: > I am thinking about how to accomplish this by spinning up multiple replicas. I guess we would need to use same `consumer-group-id` on all replicas and probably to instruct them to consumer from different partitions?

Yes that results in the same behavior but split into instances instead of threads. Using same consumer-group is the default how we deploy our services with multiple replicas - and the only one that makes sense in case of multiple replicas of the same service.

> Also, I am thinking about whether it is easier, that for the beginning - until we don't know the resource requirements, we start by having 1 replica set, and them horizontally scale if it is really needed.

In production you should always have more than one replica. In dev and test the performance should not be an issue. Just for information: We have customers who run 9 replicas of services that consume Kafka topics with high amount of data in production.

In the end, we will introduce another option we have to maintain and I just questioned, if it's worth it. I don't know the impact of performance gain vs. resource consumption. I just searched a bit and found [this information from Confluent](https://www.confluent.io/blog/kafka-consumer-multi-threaded-messaging/#multi-threaded-consumer-architecture) in regards to multi-threaded messaging for horizontal scaling:

> If you are familiar with basic Kafka concepts, you know that you can parallelize message consumption by simply adding more consumers in the same group. However, that approach is more suitable for horizontal scaling where you add new consumers by adding new application nodes (containers, VMs, and even bare metal instances).

That matches my assumption but does not tell about the performance and resource impact. Regarding the documentation and your explanation, horizontal scaling is what is covered with multiple threads provided by the used config.

The article mentions other options of multi-threaded messaging for vertical scaling. In my opinion we don't want to provide that kind of multi-threaded processing in the library - and it's not introduced with your PR. The article mentions - besides possible performance improvements - important downsides you have to accept regarding commits and message order.

vladimirsvicevicsrb:
> > > > Looks pretty straightforward to me. Are we able to add a test?
> > > 
> > > 
> > > I didn't come up with an idea on how to test that locally, as it looks like I am testing spring internals. Any idea would be appreciated. At least, I ran existing tests found in `sda-spring-boot-commons` and they pass.
> > > I prepared two PRs:
> > > 
> > > * [test: test kafka consumer concurrency configuration medical-claim-ods#63](https://github.com/SDA-SE/medical-claim-ods/pull/63)
> > > * [test: test kafka consumer concurrency configuration gitops-health-stack#269](https://github.com/SDA-SE/gitops-health-stack/pull/269)
> > >   in order to test one ODS with `PR snapshot`
> > 
> > 
> > You could create a test to see if the properties are being set, like in the test `org.sdase.commons.spring.boot.s3.config.S3ConfigurationTest`. I agree that we don’t need to test the library, so I would just add a test to check whether the config is being correctly set or not
> 
> Thanks for the hint. I will have a look into it.

Test has been added now.

sda-commons-starter-kafka/src/main/java/org/sdase/commons/spring/boot/kafka/config/KafkaConsumerConfig.java
elbiocaetano: It has, indeed, some important points when using Kafka concurrency e.g: https://docs.spring.io/spring-kafka/reference/html/#using-ConcurrentMessageListenerContainer and https://docs.spring.io/spring-kafka/reference/html/#thread-safety.
But on the other hand, we are setting the default value to 1 anyways. So it would be on the decision of the consumers to change that priority or not.
In past experiences, when we had to process a big quantity of data, was easier to increase that property and run the pods with more cpu and memory, than adding the partitions to the Kafka, because of infra restrictions.
So I would agree to approve this PR and add this functionality to spring boot commons.

JoergSiebahn: > to increase that property and run the pods with more cpu and memory, than adding the partitions to the Kafka, because of infra restrictions.

If it's creating multiple consumers independent of the actual number of partitions in the topic, I would strongly recommend not to use it. But as I understand, there will always be only one consumer per partition.

@vladimirsvicevicsrb, if I understand it correctly it would behave like the following:
Given a topic with two partitions and a service with 10 configured consumer threads, then only two of the threads will read - each from one partition - and 8 threads will idle. Is that correct? At least that would be the behavior if you have 10 replicas with the same consumer group. If 10 threads read from two topics, „all“ guarantees by Kafka are basically gone.

elbiocaetano: > If it's creating multiple consumers independent of the actual number of partitions in the topic, I would strongly recommend not to use it. But as I understand, there will always be only one consumer per partition.

Yes, this is the default behavior, unless you change the partition strategy, according to the link I sent:

> spring.kafka.consumer.properties.partition.assignment.strategy=org.apache.kafka.clients.consumer.RoundRobinAssignor

PR #207, opened 2023-03-07
Gradle 8

JoergSiebahn:


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.6-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.0.2-bin.zip
+networkTimeout=10000
```
eikenilsknopp, line 4: Is this necessary for Gradle 8 or did you run into issues?

eikenilsknopp (approved):

gradle/wrapper/gradle-wrapper.properties
JoergSiebahn: I just executed `./gradlew wrapper --gradle-version 8.0.2` (twice) and that was the result.

JoergSiebahn: Not sure, if I really want to think about the effect the property has 😄

PR #208, opened 2023-03-08
docs: adding metadata context documentation

elbiocaetano:


docs/web/index.md
```diff
@@ -539,3 +540,55 @@ Logs are printed to standard out.
...
+Most likely, this will happen when a business entity is stored based on a request and loaded later
+for further processing by a scheduler or due to a new user interaction.
+In this case, the `DetachedMetadataContext` must be persisted along with the entity and recreated
+when the entity is loaded.
+The `DetachedMetadataContext` can be defined as field in any MongoDB entity.
+
+Services that handle requests or messages in parallel must transfer the metadata context to their
+`Runnable` or `Callable` with `MetadataContext.transferMetadataContext(…)`.
+In most cases, developers should prefer`ContainerRequestContextHolder.transferRequestContext(…)`,
+which also transfers the metadata context.
```
JoergSiebahn, line 594: Imho this does not apply for Spring Boot. Or at least is in most cases not required because we should have an implementation like [this](https://github.com/SDA-SE/sda-spring-boot-commons/blob/c20a626fe9135178d7dc7a898937308203829f39/sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/async/ContextCopyTaskDecorator.java#L22) to support `@org.springframework.scheduling.annotation.Async`.

JoergSiebahn: Do we have something like that or do we need another task before we can merge the feature?

PR #211, opened 2023-03-10
refactor: own metadata context module

elbiocaetano:


sda-commons-metadata-context/src/main/java/org/sdase/commons/spring/boot/metadata/context/MetadataContextHolder.java
```diff
@@ -7,23 +7,23 @@
  */
 package org.sdase.commons.spring.boot.metadata.context;
 
-class MetadataContextHolder {
+public class MetadataContextHolder {
```
JoergSiebahn, line 10: I would not make this class and it's methods public but use `MetadataContext` in the interceptor.

sda-commons-metadata-context/build.gradle
```diff
@@ -1,15 +1,8 @@
...
   api 'org.apache.commons:commons-lang3'
-  api 'javax.ws.rs:javax.ws.rs-api'
-  api 'org.glassfish.jersey.core:jersey-common', {
-    exclude group: 'jakarta.ws.rs', module: 'jakarta.ws.rs-api'
-  }
 
   // overall testing dependencies
-  testImplementation project(':sda-commons-web-testing')
-  testImplementation project(':sda-commons-web-autoconfigure')
+  testImplementation 'org.springframework.boot:spring-boot-starter-test'
```
JoergSiebahn, line 5: I think, we can simplify the test dependencies even more:

```suggestion
  testImplementation 'org.junit.jupiter:junit-jupiter'
  testImplementation 'org.assertj:assertj-core'
```

JoergSiebahn (approved):

PR #213, opened 2023-03-10
fix(web-autoconfigure): transferring metadata context to new threads

elbiocaetano:


sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/metadata/MetadataAsyncClientAdapter.java
```diff
@@ -0,0 +1,41 @@
+/*
```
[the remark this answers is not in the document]
elbiocaetano: Good idea. I moved all the auxiliary test classes to this package

sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/metadata/test/MetadataAsyncClientAdapter.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  @Async
+  public CompletableFuture<Object> getSomethingAsync() {
+    var something = otherServiceAuthenticatedClient.getSomething();
+    var jsonNodes = objectMapper.convertValue(something, ObjectNode.class);
+    var time = System.nanoTime();
+    var timeAsJsonNode = objectMapper.convertValue(time, JsonNode.class);
+    jsonNodes.set("time", timeAsJsonNode);
+    return new AsyncResult<>(objectMapper.convertValue(jsonNodes, Object.class)).completable();
```
christopher-cudennec, line 39: Do you really have to serialize manually? Can't you just return the `jsonNodes` object?

elbiocaetano: Fixed

docs/web/index.md
```diff
@@ -577,18 +577,18 @@ Services that use the sda-spring-boot-commons:
 - are configurable by the property or environment variable `METADATA_FIELDS` to be aware of the
   metadata used in a specific environment
 
-Services that interrupt a business process should persist the context from
-`MetadataContext.detachedCurrent()` and restore it with `MetadataContext.createContext(…)` when the
-process continues.
+### Async processes
+
```
JoergSiebahn, line 581: Imho, we should keep the introduction. The first part of these paragraphs is about interrupting in terms of persisting. The last paragraph is about async (line 591+ in the old version).

docs/web/index.md
```diff
@@ -585,10 +587,11 @@ or even another instance of the service.
...
-In most cases, developers should prefer`ContainerRequestContextHolder.transferRequestContext(…)`,
-which also transfers the metadata context.
\ No newline at end of file
+when the entity is loaded. The `DetachedMetadataContext` can be defined as field in any MongoDB
+entity.
+So, for services that handle requests or messages in parallel, the
+class [ContextCopyTaskDecorator](../../sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/async/ContextCopyTaskDecorator.java)
+will copy the
+metadata context attributes to the new threads, if
+the `org.springframework.boot.autoconfigure.task.TaskExecutionAutoConfiguration` class is
```
JoergSiebahn, line 596: Do we need to mention the details and the links to the classes? As we value auto configuration and bring the neccessary dependencies, a service developer has to do nothing but add `@Async` to the method. So I would rather document that the metadata context is automatically transferred to new threads for multithreaded processing, if `@Async` is used. That is similar to [the docs about async above](https://github.com/SDA-SE/sda-spring-boot-commons/blob/a4c10a9ca4c5b86691d9485ddb5d7585732429d2/docs/web/index.md#async) - so could be even shorter.

docs/web/index.md
```diff
@@ -577,6 +577,8 @@ Services that use the sda-spring-boot-commons:
 - are configurable by the property or environment variable `METADATA_FIELDS` to be aware of the
   metadata used in a specific environment
 
+### Async processes
```
JoergSiebahn, line 580: The new headline confuses me. We have an abstract here about the metadata context for services that are not interested in it. Actually if we need more sections with subheadlines it is imho too complicated to support it.

docs/web/index.md
```diff
@@ -585,10 +587,11 @@ or even another instance of the service.
...
-The `DetachedMetadataContext` can be defined as field in any MongoDB entity.
-
-Services that handle requests or messages in parallel must transfer the metadata context to their
-`Runnable` or `Callable` with `MetadataContext.transferMetadataContext(…)`.
-In most cases, developers should prefer`ContainerRequestContextHolder.transferRequestContext(…)`,
-which also transfers the metadata context.
\ No newline at end of file
+when the entity is loaded. The `DetachedMetadataContext` can be defined as field in any MongoDB
+entity.
+So, for services that handle requests or messages in parallel, the
```
JoergSiebahn, line 592: I would keep beginning a new paragraph here as it is a different topic than persistence.

PR #221, opened 2023-03-16
fix: Sort content of openapi.yaml

christopher-cudennec:
Improving reproducible results


christopher-cudennec:
I think I prefer approach #222

PR #258, opened 2023-06-01
feat: modules cloudevents, asyncapi and yaml added

sliesche:
Not marked as **breaking change** since features were only added not changed

Notable changes to dropwizard:

- sda-commons-cloudevents: tests adapted to Spring Boot including usage of spring kafka JsonDeserializer
- sda-commons-asyncapi: uses now Spring Boot injection instead of static methods, split into separate classes, prototype scope for JsonBuilder, AsyncBuilder. Spring OOTB ObjectMapper is used


sda-commons-asyncapi/README.md
```diff
@@ -0,0 +1,143 @@
+# SDA Commons Shared AsyncAPI
```
christopher-cudennec, line 1: Oh, more than I expected. I thought about using [SpringWolf](https://github.com/springwolf/springwolf-core) for AsyncAPI documentation. I got this as another missing feature for Spring Boot Commons in the back of my head. Let's discuss this in the team.

sda-commons-cloudevents/src/test/java/org/sdase/commons/server/cloudevents/ContractCloudEventTest.java
```diff
@@ -0,0 +1,112 @@
...
+    classes = KafkaTestApp.class,
+    properties = {
+      "management.server.port=0",
+      "spring.kafka.bootstrap-servers=${spring.embedded.kafka.brokers}",
+      "spring.kafka.producer.value-serializer=org.springframework.kafka.support.serializer.JsonSerializer"
+    })
+@EmbeddedKafka(
+    partitions = 1,
+    brokerProperties = {"listeners=PLAINTEXT://localhost:9092", "port=9092"})
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
```
christopher-cudennec, line 48: Do you know the reason why we use `DirtiesContext` here? Does every Kafka test need/use that annotation?

sliesche: My understanding is that the EmbeddedKafka annotation adds an embedded Kafka to the test context which is cleaned up using DirtiesContext. used in the example provided by the spring boot kafka [documentation](https://docs.spring.io/spring-kafka/reference/html/#embedded-kafka-annotation) If not used there is a chance that some spring caching magic uses the same kafka broker in other tests.

Actually i think that both Kafka tests in cloud events could use the same broker which would speed up test execution. The documentation provides an [example](https://docs.spring.io/spring-kafka/reference/html/#same-broker-multiple-tests), although it would be easier to just put the two test in the same test class.

However, these test provide an easy to understand and safe example that will be most likely used in other tests

sda-commons-cloudevents/build.gradle
```diff
@@ -0,0 +1,12 @@
+dependencies {
+
+  api ('org.springframework.boot:spring-boot-starter-validation')
```
JoergSiebahn, line 3: Do we really need this dependency for production? I would prefer to only include `javax.validation` which is referenced in imports.

sda-commons-asyncapi/README.md
JoergSiebahn: I would agree that the AsyncAPI generation is beyond the scope of the current task. It also bloats this PR quite a lot. However, it is another dependency that is currently taken from our Dropwizard library and mixes up the dependency tree in the Spring services.

I heard mixed opinions about Spring Wolf from customers in our microservice context. In general the result seem to be _good enough_ but not completely satisfying. The AsyncAPIs we can produce with our current implementation are describing the API very well. There are still some edge cases that can't be documented as good as it should be (e.g. describing object type fields separately from the object definition). Also the library is quite old and should only be used in test code.

I think, it's not too bad to have a solution in place right away. But I would prefer further research to find a better (and actively maintained) solution. That may be Spring Wolf if we find a way to create a satisfying result that is easy to use. What do you think about marking the AsyncAPI module as a workaround alpha solution that is subject to be replaced without further notice?

sda-commons-yaml/README.md
```diff
@@ -0,0 +1,78 @@
+# SDA Commons Shared YAML
```
JoergSiebahn, line 1: Do we really need this module? Isn't there something available from Spring? I would like to avoid maintaining more code than necessary.

sda-commons-asyncapi/README.md
sliesche: Yaml module removed.

Updated Async Readme

> ## ⚠️ Experimental ⚠️
>
> ### Please be aware that SDA SE is likely to change or remove this artifact in the future

sda-commons-yaml/README.md
sliesche: module removed

`com.fasterxml.jackson.dataformat.yaml.YAMLMapper` used in asyncapi module

sda-commons-asyncapi/src/test/java/org/sdase/commons/spring/boot/asyncapi/internal/JsonSchemaEmbedderTest.java
```diff
@@ -0,0 +1,75 @@
...
+  //
+  //    assertThat(result.at("/definitions").fieldNames())
+  //        .toIterable()
+  //        .containsExactlyInAnyOrder("Person", "simpleobjectsimplereferencedyaml");
+  //    assertThat(result.at("/definitions/simpleobjectsimplereferencedyaml").isMissingNode())
+  //        .isFalse();
+  //    assertThat(result.at("/definitions/address").isMissingNode()).isTrue();
+  //    assertThat(result.at("/definitions/Person/properties/address/$ref").asText())
+  //        .isEqualTo("https://sda.se");
+  //  }
```
JoergSiebahn, line 74: NIT: Could we remove this code, enable the test or disable the test?

sda-commons-cloudevents/README.md
```diff
@@ -0,0 +1,63 @@
...
+> The bundle is still in an early state. APIs are open for change.
+>
+
+## Introduction
+
+CloudEvents is a general standard that can be used in combination with your favourite eventing tool
+like ActiveMQ or Kafka. The CloudEvents specification defines concrete bindings to define how the
+general specification should be applied to a specific tool. This module brings you
+the [Kafka protocol binding](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/bindings/kafka-protocol-binding.md)
+.
```
JoergSiebahn, line 16: NIT

```suggestion
the [Kafka protocol binding](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/bindings/kafka-protocol-binding.md).
```

JoergSiebahn (approved):

sda-commons-cloudevents/README.md
sliesche: updated

sda-commons-asyncapi/src/test/java/org/sdase/commons/spring/boot/asyncapi/internal/JsonSchemaEmbedderTest.java
sliesche: good find! i deactivated it to fix something

enabled test

JoergSiebahn (approved):

PR #265, opened 2023-06-14
fix: instantiation of clientRegistrationRepository

elbiocaetano:


sda-commons-web-testing/src/main/java/org/sdase/commons/spring/boot/web/testing/auth/AuthMock.java
```diff
@@ -29,9 +29,9 @@ public class AuthMock {
 
   private static final Logger LOG = LoggerFactory.getLogger(AuthMock.class);
 
-  private static final String DISCOVERY_PATH = "/issuer/.well-known/openid-configuration";
-  private static final String JWKS_PATH = "/issuer/keys";
-  private static final String TOKEN_PATH = "/issuer/token";
+  public static final String DISCOVERY_PATH = "/issuer/.well-known/openid-configuration";
+  public static final String JWKS_PATH = "/issuer/keys";
+  public static final String TOKEN_PATH = "/issuer/token";
```
JoergSiebahn, line 34: I would not use the reference to „production“ properties in test verification. If you just change the values here, the tests will not fail, although the API changes.

```suggestion
  private static final String DISCOVERY_PATH = "/issuer/.well-known/openid-configuration";
  private static final String JWKS_PATH = "/issuer/keys";
  private static final String TOKEN_PATH = "/issuer/token";
```

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/client/SdaOidcClientConfiguration.java
```diff
@@ -9,41 +9,63 @@
...
-   * @return the client registration repository containing the oidc client registration.
+   * @return oAuth2TokenProvider
    */
-  @Bean
-  InMemoryClientRegistrationRepository clientRegistrationRepository(
-      @Value("${oidc.client.id}") String clientId,
-      @Value("${oidc.client.secret}") String clientSecret,
-      @Value("${oidc.client.issuer.uri}") String oidcIssuer) {
+  @Nullable
+  public OAuth2TokenProvider oAuth2Provider() {
```
sliesche, line 56: Seems like you implementing a manual lazy loading.

It should be possible to annotate a Bean with` @Lazy` and then use `applicationContext.getBean(OAuth2TokenProvider.class);`  in the `RequestInterceptor`to control when the bean is instantiated the first time

sliesche (approved):

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/client/SdaOidcClientConfiguration.java
```diff
@@ -10,40 +10,58 @@
...
-  InMemoryClientRegistrationRepository clientRegistrationRepository(
-      @Value("${oidc.client.id}") String clientId,
-      @Value("${oidc.client.secret}") String clientSecret,
-      @Value("${oidc.client.issuer.uri}") String oidcIssuer) {
+  @Lazy
+  public OAuth2TokenProvider oAuth2Provider() throws IllegalArgumentException {
+    return new OAuth2TokenProvider(authorizedClientManager());
+  }
+
+  public OAuth2AuthorizedClientManager authorizedClientManager() {
```
vladimirsvicevicsrb, line 55: I guess that all methods `line: 55 - 62` can be `private`

sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/client/ClientCredentialsFlowIT.java
```diff
@@ -77,12 +103,127 @@ void shouldUseAccessTokenFromRequestContext() {
...
+    // given
+    authMock.reset();
+    authMock.authorizeAnyRequest().allow();
+
+    // when
+    request.getForObject(
+        String.format("http://localhost:%d/api/ping/external", port), Object.class);
+
+    // then
+    authMock.wireMockServer().verify(2, getRequestedFor(urlPathEqualTo(DISCOVERY_PATH)));
```
vladimirsvicevicsrb, line 185: Just wondering: why do we have calls again here instead of using cached values like [here](https://github.com/SDA-SE/sda-spring-boot-commons/pull/265/files#diff-ed3b61b35b9860e7806d4c9eeb18aa6a62ca62b833282d8e37064bcb6888d458R227)? 🤔

elbiocaetano: Because for this one the first request failed, so it must send a request to oidc issuer url again, until it retrieves the data successfully, that are the other scenarios

vladimirsvicevicsrb (approved):

PR #289, opened 2023-08-01
fix(s3): temporary quick fix for S3 compile errors

JoergSiebahn:


JoergSiebahn:
closed in favor of #290

PR #291, opened 2023-08-01
WIP feat: update spring security

eikenilsknopp:


sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/auth/SdaAccessDecisionManager.java
```diff
@@ -8,19 +8,36 @@
...
-public class SdaAccessDecisionManager extends UnanimousBased {
+public class SdaAccessDecisionManager implements AuthorizationManager<FilterInvocation> {
+  private AuthorizationManager<FilterInvocation> authorizationManager;
 
   public SdaAccessDecisionManager(
       ManagementAccessDecisionVoter managementAccessDecisionVoter,
       OpaExcludesDecisionVoter opaExcludesDecisionVoter,
       OpaAccessDecisionVoter opaAccessDecisionVoter) {
-    super(List.of(managementAccessDecisionVoter, opaExcludesDecisionVoter, opaAccessDecisionVoter));
+    authorizationManager = AuthorizationManagers.allOf(managementAccessDecisionVoter, opaExcludesDecisionVoter, opaAccessDecisionVoter);
```
JoergSiebahn, line 31: Imho this must be `anyOf(…)`. In this previous implementation we always used `ACCESS_ABSTAIN` instead of deny. It means, that the voter returning it, does not decide for this request. The aggregation rule to allow a request was „_at minimum one granted and no deny decision, ignoring any abstain decisions_“. As we did not use deny explicitly, a request was denied, when all voters returned abstain.

sda-commons-app-example/openapi.yaml
```diff
@@ -2,6 +2,9 @@ openapi: 3.0.1
 info:
   title: OpenAPI definition
   version: v0
+servers:
+- url: http://localhost:52960/api
+  description: Generated server url
```
eikenilsknopp, line 7: How did that end up here? 🤔

sda-commons-web-autoconfigure/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
eikenilsknopp, line 1: Feature flag to enable mocking of final classes

sda-commons-app-example/openapi.yaml
JoergSiebahn: Maybe updated by a test?

sda-commons-web-autoconfigure/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
JoergSiebahn: Is it because of `RequestAuthorizationContext`? It has a public constructor, so you may just mock the `HttpServletRequest` and create a real instance.

sda-commons-app-example/openapi.yaml
JoergSiebahn: I created a subtask. We can investigate that later.

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/auth/SdaSecurityConfiguration.java
```diff
@@ -83,17 +84,29 @@ AuthenticationManager authenticationManager(AuthenticationManagerBuilder auth) {
...
+       .authorizeRequests(
+           authorize ->
+               authorize.anyRequest().permitAll().accessDecisionManager(sdaAccessDecisionManager))
+       .oauth2ResourceServer(
+           oauth2 -> oauth2.authenticationManagerResolver(authenticationManagerResolver))
+       .headers(
+           configurer ->
+               configurer.addHeaderWriter(
+                   new StaticHeadersWriter(sdaSecurityHeaders.getSecurityHeaders())));
+    */
```
JoergSiebahn, line 109: Is there a reason to remove the csrf config?
Imho CRSF protection (with a token as in this implementation) does not make sense for stateless REST APIs

```suggestion
                    new StaticHeadersWriter(sdaSecurityHeaders.getSecurityHeaders())))
        .csrf()
        .disable();
```

sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/auth/AuthenticationIT.java
```diff
@@ -135,7 +135,7 @@ void shouldSendDataForAuthorizationToOpa() {
                         "$.input.headers.authorization[0]",
                         equalTo("Bearer " + authentication.token())))
                 .withRequestBody(
-                    matchingJsonPath("$.input.headers.connection[0]", equalTo("Keep-Alive"))));
+                    matchingJsonPath("$.input.headers.connection[0]", equalTo("keep-alive"))));
```
JoergSiebahn, line 138: NIT: Imho technically correct and more sustainable would be to ignore the case instead of adjusting the expectation:

```suggestion
                    matchingJsonPath("$.input.headers.connection[0]", equalToIgnoreCase("Keep-Alive"))));
```

JoergSiebahn (approved):
🚀

PR #296, opened 2023-08-03
fix: remove servers from generated openapi

eikenilsknopp:


sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/docs/SdaOpenApiCustomizerConfiguration.java
```diff
@@ -24,6 +25,11 @@ public OpenApiCustomizer removeServers() {
     return openApi -> openApi.servers(null);
   }
 
+  @Bean
```
JoergSiebahn, line 28: Can we remove the bean above then?

eikenilsknopp: Yes we can, removed

JoergSiebahn (approved):

PR #297, opened 2023-08-03
Fix CVEs for Spring Boot 3

JoergSiebahn:


sda-commons-dependencies/build.gradle
```diff
@@ -42,8 +46,11 @@ dependencies {
     api 'io.swagger.core.v3:swagger-annotations:2.2.9'
     api 'io.github.classgraph:classgraph:4.8.161'
     api 'javax.ws.rs:javax.ws.rs-api:2.1.1'
-    api 'org.yaml:snakeyaml:1.33', {
-      because 'Temporary: this version is used to solve CVE-2022-25857'
+    api 'org.yaml:snakeyaml:2.0', {
+      because 'fix CVE-2022-25857'
+    }
```
JoergSiebahn, line 51: The next Spring Boot Minor Version 3.2.x [will upgrade Snakeyaml](https://github.com/spring-projects/spring-boot/commit/1f9ce508f7f996bcba8d7707a06ab0210ec5089a) for us. The [deleted test](https://github.com/spring-projects/spring-boot/commit/1f9ce508f7f996bcba8d7707a06ab0210ec5089a#diff-2623de9e0a020508fffd1009a0f9bb970ae1cb017deb523c480119b51c7f1a64) indicates, that Spring itself should not have an issue with the new version. According to [this comment](https://github.com/spring-projects/spring-boot/issues/34405#issuecomment-1456219257) it should be safe if we only use `application.properties`. However, I've seen [some closed issues](https://github.com/spring-projects/spring-boot/issues?q=is%3Aissue+snakeyaml) of (unclear) transitive libraries or custom code of the issuer that failed. Should we try to force the update and see what happens or should we wait?

preethivenkatesh-sdase (approved):

sda-commons-dependencies/build.gradle
elbiocaetano: Since we are busy with the migration, I would just accept this and then when renovate update to the new version we can check if the problem still exists

JoergSiebahn: @preethivenkatesh-sdase fyi

PR #301, opened 2023-08-04
feat: migrating to open telemetry

elbiocaetano:


sda-commons-web-autoconfigure/src/main/resources/org/sdase/commons/spring/boot/web/defaults.properties
```diff
@@ -6,3 +6,9 @@ server.error.whitelabel.enabled=false
 server.tomcat.basedir=${java.io.tmpdir}/tomcat
 
 request.body.max.size=1 MB
+
```
JoergSiebahn, line 9: Should this additional properties be moved to `tracing.properties`/`monitoring.properties`?

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/monitoring/SdaTracingConfiguration.java
```diff
@@ -7,9 +7,22 @@
...
 @PropertySource("classpath:/org/sdase/commons/spring/boot/web/monitoring/tracing.properties")
 @AutoConfiguration
-public class SdaTracingConfiguration {}
+public class SdaTracingConfiguration {
+
+  @Bean
+  @ConditionalOnProperty(
+      value = "management.otlp.disable",
+      havingValue = "true",
+      matchIfMissing = true)
```
JoergSiebahn, line 24: If `matchIfMissing` is `true`, a would expect that the default is the `noop` tracer. But shouldn't we have tracing enabled by default?

sda-commons-web-autoconfigure/src/main/resources/org/sdase/commons/spring/boot/web/defaults.properties
```diff
@@ -6,3 +6,9 @@ server.error.whitelabel.enabled=false
...
+management.tracing.sampling.probability=1.0
```
JoergSiebahn, line 10: Does this mean, that we propagate every trace? I'm not sure if that could cause too much tracing traffic. Do you know the previous default and the default in Dropwizard Commons?

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/auth/opa/OpaAuthorizationManager.java
```diff
@@ -79,11 +77,11 @@ public OpaAuthorizationManager(
       @Value("${opa.policy.package:}") String policyPackage,
       OpaRequestBuilder opaRequestBuilder,
       @Qualifier("opaRestTemplate") RestTemplate opaRestTemplate,
-      ApplicationContext applicationContext
-      /*, TODO update to opentelemetry Object tracer*/ ) {
+      ApplicationContext applicationContext,
+      OpenTelemetry openTelemetry) {
```
JoergSiebahn, line 81: NIT: We removed the `@param` in the JavaDoc when it was not available. Can you add the docs again?

sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/auth/opa/OpaAuthorizationManagerTest.java
```diff
@@ -106,15 +119,17 @@ void shouldUseConfiguredPolicyPackage() {
...
     assertThat(decisionVoter)
         .extracting("opaRequestUrl")
         .asString()
         .isEqualTo(opaBaseUrl + "/v1/data/com/example/package");
   }
 
-  @Test
+  @Test // Request {method=POST,
+  // url=http://localhost:4317/opentelemetry.proto.collector.trace.v1.TraceService/Export,
+  // headers=[User-Agent:OTel-OTLP-Exporter-Java/1.28.0, te:trailers]}
```
JoergSiebahn, line 132: Something that needs to be removed? 😉

sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/monitoring/TracingFeignClientIT.java
```diff
@@ -43,18 +48,32 @@
     webEnvironment = WebEnvironment.RANDOM_PORT,
     properties = {
       "test.tracing.client.base.url=http://localhost:${wiremock.server.port}/feign",
-      "spring.zipkin.base-url=http://localhost:${wiremock.server.port}/zipkin",
-      "opa.disable=true"
+      "otel.traces.exporter=zipkin",
+      "otel.exporter.zipkin.endpoint=http://localhost:${wiremock.server.port}/zipkin/api/v2/spans",
+      "opa.disable=true",
+      "otel.metrics.exporter=none",
+      "otel.logs.exporter=none",
```
JoergSiebahn, line 55: Should this be default properties in the lib?

JoergSiebahn: Seems like we need to double check, what's really needed here and if more needs to be part of the default config.

sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/monitoring/TracingFeignClientIT.java
```diff
@@ -103,7 +122,7 @@ void shouldExtractTraceAndSpanForB3ContextPropagationAndUseInClient() {
         client.exchange(
             RequestEntity.get(
                     URI.create(String.format("http://localhost:%d/api/pingMetrics", port)))
-                .header("x-b3-traceid", "0000000000000001")
+                .header("X-B3-TraceId", "0000000000000001")
```
JoergSiebahn, line 125: Should it work with any case? Then we maybe need to duplicate the test method.

sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/monitoring/TracingFeignClientIT.java
```diff
@@ -89,7 +108,7 @@ void shouldExtractTraceAndSpanForW3CContextPropagationAndUseInClient() {
 
     verify(
         postRequestedFor(urlPathEqualTo("/zipkin/api/v2/spans"))
-            .withRequestBody(containing("\"traceId\":\"0000000000000001\"")));
+            .withRequestBody(new AnythingPattern()));
```
JoergSiebahn, line 111: Should we be more specific as before?

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/monitoring/SdaTracingConfiguration.java
elbiocaetano: I agree, but for the tests it must be disabled by default. I added to the test system properties [here](https://github.com/SDA-SE/sda-spring-boot-commons/pull/301/files#diff-943d09cac58ce340d7fb07e57228f6feaf0d8513e7cf277cd02de523284ce0f1R41)

sda-commons-web-autoconfigure/src/main/resources/org/sdase/commons/spring/boot/web/defaults.properties
elbiocaetano: dw commons uses 1.0 as default from class `io.opentelemetry.sdk.autoconfigure.TracerProviderConfiguration` property `DEFAULT_TRACEIDRATIO_SAMPLE_RATIO = 1.0d;`

sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/monitoring/TracingFeignClientIT.java
elbiocaetano: Yes, the correct should be lower case, I fixed it

sda-commons-web-autoconfigure/src/test/java/org/sdase/commons/spring/boot/web/monitoring/TracingFeignClientIT.java
elbiocaetano: The old format was zipkin, now it is using the otlp format, so the content changed. The body from otlp does not have the trace id, so I changed the assertion for something that should always be there

sda-commons-web-autoconfigure/src/main/java/org/sdase/commons/spring/boot/web/monitoring/SdaTracingConfiguration.java
JoergSiebahn: Does that apply for users of the library as well?
Would it make sense to disable it in `src/test/resources/application.properties` instead of the gradle file?

eikenilsknopp (approved):

PR #304, opened 2023-08-08
generate async api without outdated tools for v3

JoergSiebahn:
Kudos to @Fox32. This implementation is inspired by https://github.com/SDA-SE/sda-dropwizard-commons/pull/409


.github/workflows/temp-java-ci-upgrade.yaml
```diff
@@ -52,3 +52,8 @@ jobs:
             :sda-commons-starter-s3:test
             :sda-commons-starter-web:test
             :sda-commons-web-testing:test
+            :sda-commons-cloudevents:test
```
JoergSiebahn, line 55: Seems like the regular build (Windows still running) is green now. Should we remove this temporary workflow and rely on the regular checks?

PR #307, opened 2023-08-09
fix: set default properties for mongodb tests

eikenilsknopp:


sda-commons-starter-mongodb/src/main/resources/org/sdase/commons/spring/boot/mongodb/defaults.properties
```diff
@@ -1 +1 @@
-
+de.flapdoodle.mongodb.embedded.version=4.4.1
```
elbiocaetano, line 1: Isn't this used only for tests? Shouldn't it be in the test application.properties file?

elbiocaetano (approved):

PR #324, opened 2023-08-23
Plp 1125

frederic-huebner:


sda-commons-starter-kafka/src/test/java/org/sdase/commons/spring/boot/kafka/KafkaRetryAndDltConsumerTest.java
```diff
@@ -44,7 +44,10 @@
 @SpringBootTest(
     classes = KafkaTestApp.class,
     webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
-    properties = {"spring.kafka.bootstrap-servers=${spring.embedded.kafka.brokers}"})
+    properties = {
+      "spring.kafka.bootstrap-servers=${spring.embedded.kafka.brokers}",
+      "sda.kafka.consumer.dlt.pattern="
```
JoergSiebahn, line 49: Is there also an integration test that verifies the default and still has the pattern set to `dlt-<topic>`? I would have expected some diff for the opposite of this test where the setting is removed and the test still works.

JoergSiebahn (approved):

PR #338, opened 2023-08-31
Add test support for external MongoDB

JoergSiebahn:
- [x] test in real service with real pipeline when snapshots are pushed to Nexus

### DIFF
--- a/.github/workflows/mongodb.yml
+++ b/.github/workflows/mongodb.yml
@@ -25,7 +25,7 @@ jobs:
           cache: 'gradle'
 
       - name: Start MongoDB
-        run: docker run -d -e MONGO_INITDB_ROOT_USERNAME=test -e MONGO_INITDB_ROOT_PASSWORD=example -p 27017:27017 --name test_mongo mongo:4.4
+        run: docker run -d -e MONGO_INITDB_ROOT_USERNAME=test -e MONGO_INITDB_ROOT_PASSWORD=example -p 27017:27017 --name test_mongo mongo:5.0
 
       - name: Test with Gradle
         run: ./gradlew :sda-commons-starter-mongodb:test
--- a/.github/workflows/mongodb.yml
+++ b/.github/workflows/mongodb.yml
@@ -1,0 +1,38 @@
+# This workflow tests the feature of using an external MongoDB for tests
+name: Java Test MongoDB
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  # Test with real MongoDB
+  test-mongodb:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    env:
+      TEST_MONGODB_CONNECTION_STRING: mongodb://test:example@localhost:27017/testdb?authSource=admin
+    steps:
+      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@5ffc13f4174014e2d4d4572b3d74c3fa61aeb2c2 # v3.11.0
+        with:
+          distribution: 'temurin'
+          java-version: 17
+          cache: 'gradle'
+
+      - name: Start MongoDB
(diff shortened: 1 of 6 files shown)


.github/workflows/mongodb.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@5ffc13f4174014e2d4d4572b3d74c3fa61aeb2c2 # v3.11.0
+        with:
+          distribution: 'temurin'
+          java-version: 17
+          cache: 'gradle'
+
+      - name: Start MongoDB
+        run: docker run -d -e MONGO_INITDB_ROOT_USERNAME=test -e MONGO_INITDB_ROOT_PASSWORD=example -p 27017:27017 --name test_mongo mongo:4.4
```
christopher-cudennec, line 28: We could go to MongoDB 5 like in dw-commons.

sda-commons-web-testing/src/main/java/org/sdase/commons/spring/boot/web/testing/mongodb/ExternalMongoAutoConfiguration.java
```diff
@@ -0,0 +1,77 @@
...
+import org.springframework.core.env.ConfigurableEnvironment;
+import org.springframework.core.env.MapPropertySource;
+
+public class ExternalMongoAutoConfiguration implements EnvironmentPostProcessor {
+  private static final String FLAPDOODLE_AUTO_CONFIG_CLASS =
+      "de.flapdoodle.embed.mongo.spring.autoconfigure.EmbeddedMongoAutoConfiguration";
+  private static final String TEST_MONGODB_CONNECTION_STRING_PROPERTY_NAME =
+      "test.mongodb.connection.string";
+  private static final String AUTOCONFIGURE_EXCLUDES_PROPERTY_NAME = "spring.autoconfigure.exclude";
+  private static final String SPRING_DATA_MONGODB_URI_PROPERTY_NAME = "spring.data.mongodb.uri";
```
christopher-cudennec, line 25: Can we make the same approach without introducing our own property? Can't we implement the same behaviour if we pass `SPRING_DATA_MONGODB_URI` from outside?

.github/workflows/mongodb.yml
JoergSiebahn: Funny … I had that before and switched back because I found 4.4 in the task. Thought it has a reason.

sda-commons-web-testing/src/main/java/org/sdase/commons/spring/boot/web/testing/mongodb/ExternalMongoAutoConfiguration.java
JoergSiebahn: It is by intention to use the same environment variable as [in Dropwizard Commons](https://github.com/SDA-SE/sda-dropwizard-commons/blob/master/sda-commons-server-mongo-testing/src/main/java/org/sdase/commons/server/mongo/testing/MongoDb.java#L43). That's also the default in our build pipelines. We would require the same code because Flapdoodle's auto configuration does not care about any settings you apply.

JoergSiebahn: And the environment variable is also named in the task 😉

.github/workflows/mongodb.yml
christopher-cudennec: I would just use the same setup like in DW. So let's update to 5 since this is [now](https://github.com/SDA-SE/sda-dropwizard-commons/commit/d1c7b6d8abf258f58659b1bc73c6385c4ce7cba6) the version that we use there.

sda-commons-web-testing/src/main/java/org/sdase/commons/spring/boot/web/testing/mongodb/ExternalMongoAutoConfiguration.java
christopher-cudennec: Ok. The Jenkins Pipeline will work out-of-the-box in this case. Let's keep it!

.github/workflows/mongodb.yml
JoergSiebahn: bumped to 5.0

christopher-cudennec (approved):

PR #342, opened 2023-09-04
Docs migration guide: parameters compile task

JoergSiebahn:

PR #343, opened 2023-09-04
add workflow to publish GH pages

JoergSiebahn:
see #342

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -1,0 +1,42 @@
+name: Publish mkdocs
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+jobs:
+  deploy:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b #v3.0.2
+        with:
+          fetch-depth: 1
+      - run: git config user.name 'github-actions[bot]' && git config user.email 'github-actions[bot]@users.noreply.github.com'
+      - run: sudo apt-get update && sudo apt-get -yq install graphviz git fonts-ipafont fonts-ipaexfont
+      - run: >
+          wget -q -O /usr/local/bin/plantuml https://github.com/mikaelhg/puni2/releases/download/1.2022.7/plantuml-1.2022.7-glibc-x86_64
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,42 @@
...
+  deploy:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b #v3.0.2
+        with:
+          fetch-depth: 1
+      - run: git config user.name 'github-actions[bot]' && git config user.email 'github-actions[bot]@users.noreply.github.com'
+      - run: sudo apt-get update && sudo apt-get -yq install graphviz git fonts-ipafont fonts-ipaexfont
+      - run: >
+          wget -q -O /usr/local/bin/plantuml https://github.com/mikaelhg/puni2/releases/download/1.2022.7/plantuml-1.2022.7-glibc-x86_64
```
christopher-cudennec, line 17: I'm fine with using this version as I expect it is already used + tested somewhere else. But also please consider updating at some point the future as there are many [more recent releases](https://github.com/mikaelhg/puni2/releases).

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,42 @@
...
+      - run: pip install mkdocs-techdocs-core # Includes material theme
+      - run: pip install mkdocs-build-plantuml-plugin
+      - run: git fetch origin gh-pages # Get latest history
+      - run: > # required for mkdocs, not for techdocs
+          yq -i '.plugins |= . +
+          [{"build_plantuml": {"render":"server", "server":"http://www.plantuml.com/plantuml", "output_format": "svg"}}]'
+          mkdocs.yaml
+      - name: Dry run generated mkdocs
+        if: github.event_name == 'pull_request'
+        run: mkdocs build --config-file mkdocs.yaml && ls -al
```
christopher-cudennec, line 31: `-v` could help debugging in case of problems

christopher-cudennec (approved):

.github/workflows/publish-docs.yml
JoergSiebahn: I assume that has been aligned with the version in Backstage. Not sure if that is still the case though.

PR #356, opened 2023-09-18
fix(asyncapi): merge the same type if generated twice

JoergSiebahn:


sda-commons-asyncapi/src/main/java/org/sdase/commons/spring/boot/asyncapi/jsonschema/JsonSchemaBuilder.java
```diff
@@ -77,6 +78,9 @@ default Map<String, JsonNode> toJsonSchema(Iterable<Type> types) {
                 Map.Entry::getKey,
                 Map.Entry::getValue,
                 (v1, v2) -> {
+                  if (Objects.equals(v1, v2)) {
```
elbiocaetano, line 81: We can also use .distinct() in the stream e.g:

 ```java
 return StreamSupport.stream(types.spliterator(), false)
        .distinct()
        .map(this::toJsonSchema)
        .map(Map::entrySet)
        .flatMap(Set::stream)
        .distinct()
        .collect(...
```

elbiocaetano (approved):

sda-commons-asyncapi/src/main/java/org/sdase/commons/spring/boot/asyncapi/jsonschema/JsonSchemaBuilder.java
JoergSiebahn: Changed the implementation and added a comment.

elbiocaetano (approved):

PR #365, opened 2023-09-25
feat: not publishing releases in out internal nexus repository

elbiocaetano:
BREAKING CHANGE: The sda-spring-boot-commons releases will not be available in SDA-SE internal Nexus anymore, so you must retrieve them from Maven Central, from now on.


christopher-cudennec (commented):
Could you double-check if we need a snapshot repo for the `mavenCentral` repository (`https://oss.sonatype.org/content/repositories/snapshots`)? I only [found a GH Actions workflow](https://github.com/SDA-SE/sda-spring-boot-commons/blob/5e9d9a366ee8490a7ce9e1585421a8a5763c400b/.github/workflows/pull-request-snapshots.yml) that published snapshots to our internal Nexus.

PR #413, opened 2023-10-11
fix: always exclude apiDocsUrls from opa

eikenilsknopp:


sda-commons-starter-web/src/main/java/org/sdase/commons/spring/boot/web/auth/opa/OpaExcludesAuthorizationManager.java
```diff
@@ -31,16 +31,18 @@ public class OpaExcludesAuthorizationManager
...
           .filter(Objects::nonNull)
           .filter(p -> !p.isBlank())
           .map(String::trim)
           .map(Pattern::compile)
           .forEach(this.excludedPathPatterns::add);
     }
+
+    // Always add the apiDocsUrls
+    this.excludedPathPatterns.add(Pattern.compile(apiDocsUrl));
+    this.excludedPathPatterns.add(Pattern.compile(apiDocsUrl + ".yaml"));
```
christopher-cudennec, line 45: Could you inject the value instead by using referencing `Constant.DEFAULT_API_DOCS_URL_YAML`? We would use the same approach like for `API_DOCS_URL`.

christopher-cudennec (commented):
Can you also fix the [documentation](https://github.com/SDA-SE/sda-spring-boot-commons/blob/46edf85076d03118f61143c49d4838bf6245a22b/docs/starter-web.md?plain=1#L15) for `opa.exclude.patterns`? Something like "... `openapi` and `openapi.yaml` are always excluded".

christopher-cudennec (approved):

PR #422, opened 2023-10-18
fix: inject configured objectMapper into kafka producer

eikenilsknopp:


sda-commons-starter-kafka/src/test/java/org/sdase/commons/spring/boot/kafka/ProducerObjectMapperTest.java
```diff
@@ -0,0 +1,78 @@
...
+            () ->
+                assertThat(
+                        KafkaTestUtil.getNextRecord(
+                                topic,
+                                KafkaTestUtil.createTestConsumer(
+                                    topic, embeddedKafkaBroker, new StringDeserializer()))
+                            .value()
+                            .toString())
+                    .contains(dateString));
+  }
```
christopher-cudennec, line 77: Won't the test execute faster if you only wait for the record to be produced in `untilAsserted` and then compare the expected value after the `await()` block? Imagine that the serialization is wrong. The `untilAsserted` block would test over and over again until it runs into timeout.

sda-commons-starter-kafka/src/main/java/org/sdase/commons/spring/boot/kafka/SdaKafkaProducerConfiguration.java
```diff
@@ -27,9 +30,17 @@ public class SdaKafkaProducerConfiguration {
...
+
+  public SdaKafkaProducerConfiguration(ObjectMapper objectMapper) {
+    this.objectMapper = objectMapper;
+  }
+
   @Bean
   @SuppressWarnings("java:S1452")
   public KafkaTemplate<String, ?> kafkaTemplate(ProducerFactory<String, ?> producerFactory) {
+    ((DefaultKafkaProducerFactory<?, ?>) producerFactory)
+        .setValueSerializer(new JsonSerializer<>(objectMapper));
```
christopher-cudennec, line 43: I have a bad feeling if we substitute the value serializer by a brand new instance. What if it was not a `JsonSerializer` at all? What if it was configured in a specific way?

Have you compared the approach mentioned in the ticket to use a `DefaultKafkaProducerFactoryCustomizer`? Do you have arguments for this approach compared to the mentioned approach?

eikenilsknopp: I'm pretty sure you would need to create your own `kafkaTemplate` bean anyways if you want to configure something, so I would assume it's ok to configure this default kafkaTemplate with this JsonSerializer

christopher-cudennec: I think this is the line of code where I've seen someone setting the value serializer before: https://github.com/SDA-SE/sda-spring-boot-commons/blob/master/sda-commons-starter-kafka/src/main/resources/org/sdase/commons/spring/boot/kafka/producer.properties#L2

Will users of the library override the value?
On the other hand: As you mentioned, with Spring's flexibility you should still be able to define your own KafkaTemplate and use a different serializer.

sda-commons-starter-kafka/src/test/java/org/sdase/commons/spring/boot/kafka/ProducerObjectMapperTest.java
```diff
@@ -0,0 +1,71 @@
...
+            .setCheckString("CHECK")
+            .setOffsetDateTime(OffsetDateTime.parse(dateString));
+    testee.send(expectedMessage);
+
+    var kafkaMessage =
+        KafkaTestUtil.getNextRecord(
+                topic,
+                KafkaTestUtil.createTestConsumer(
+                    topic, embeddedKafkaBroker, new StringDeserializer()))
+            .value();
```
christopher-cudennec, line 67: Oh, you've removed the "untilAsserted" part completely. Won't the test now be unstable due to timing issues? I think it's still a good idea to wrap `KafkaTestUtil.getNextRecord` with an `await` until the retrieved value is not `null`.

sda-commons-starter-kafka/src/main/java/org/sdase/commons/spring/boot/kafka/SdaKafkaProducerConfiguration.java
sliesche: Instead of configuring Kafka to use a specific object mapper. why don't we configure the default object mapper correctly? this would have the benefit that object mapper throughout the application has the same configuration and we don't need to work with different mappers in different modules

https://www.baeldung.com/spring-boot-customize-jackson-objectmapper

Depending on the specific time format it might also be possible to just adjust properties e.g.
`spring.jackson.date-format=`
`spring.jackson.joda-date-time-format=`
`spring.jackson.serialization.write-dates-as-timestamps=false`

https://docs.spring.io/spring-boot/docs/1.4.x/reference/html/common-application-properties.html

sda-commons-starter-kafka/src/test/java/org/sdase/commons/spring/boot/kafka/ProducerObjectMapperTest.java
eikenilsknopp: `getNextRecord` uses the spring KafkaTestUtils class beneath which already polls kafka with a timeout of 60s. So as long as the kafka message doesn’t take longer than 60s this should be stable without untilAsserted.

eikenilsknopp:
I now considered 3 possible solutions and I still think that my solution in this PR is the 'least intrusive' option:

- Solution 1: Configure a DefaultKafkaProducerFactoryCustomizer with a JsonSerializer and the correct ObjectMapper
Configuring the Serializer an a ProducerFactory level applies it to all producers that are created with that factory, so it must be 'un-configured' for all producers that should use a different serializer, e.g. the DLT producer. It also needs to be unconfigured whenever a user of the library creates its own kafkaTemplate und wants to use a different serializer.

- Solution 2: Configuring the default kafkaTemplate with a JsonSerializer and the correct ObjectMapper
Defines the serializer only for this producer, which is intended as a Json producer anyway. But it is then fixed to that, if a dev wants to use a different serializer he needs to create a new Bean and can’t simply configure the `spring.kafka.producer.value-serializer` property.

- Solution 3: Configure the ObjectMapper that is used by the default JsonSerializer
Unfortunately not possible, creating a serializer implicitly (e.g. through the producer.properties) creates a new ObjectMapper from scratch without the needed configuration. See https://github.com/spring-projects/spring-kafka/blob/a455eed0843597a9fe54e04895c0c19bd98395fd/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonSerializer.java#L84

Therefore I propose we go with the current solution, solution 2.

sliesche (approved):

PR #425, opened 2023-10-19
test(asyncapi): add test to document limitations of `@NotBlank`

JoergSiebahn:


christopher-cudennec (commented):
I've already looked into it yesterday, but I must admit the setup and the benefit was not clear to me. I think it's hard to understand how the new test documents the limitations. I would've expected an expected `asyncapi.yaml` and a disabled test for `@NonBlank`. At the moment it's hidden in the implementation details (and the inheritance hierarchy of the abstract/concrete) classes?

Do you have an idea how we can simplify the setup?

JoergSiebahn:
> Do you have an idea how we can simplify the setup?

Actually, no.

> I would've expected an expected asyncapi.yaml …

1. The test is only testing the JSON Schema generator, not the AsyncAPI generator.
2. I suffered a lot, when changing the used JSON Schema library, because we only had tests that compared complete AsyncAPIs. It was completely hidden, what is actually part of the test, there was no option to express which details are under test.

To document the requirements more precisely, there is this abstract test with the parameters that define exactly one property and it's expected value for a given annotated Java representation of the model. The test is abstract and only extended by the test for Victools to be able to try out other libraries in the future and compare exactly what they support and what they don't support compared to the currently used library.

> … and a disabled test for `@NonBlank`

Not everything of the `@NotBlank` annotation is wrong. Only the `pattern` is missing. With the detailed setup, we can verify that what already worked is still green and can disable the one thing that does not work for the current implementation. That's _„how the new test documents the limitations“_ of the currently used library.

christopher-cudennec (approved):

PR #555, opened 2024-01-16
fix: application should start without definition of `management.serve…

christopher-cudennec:
…r.port`

We first need a discussion about our auto configuration approach. My feeling (I will write down some arguments) is that we should not use it to bootstrap our application.


christopher-cudennec:
closing: preferred solution was PR #556

PR #655, opened 2024-03-19
fix(deps): manage spring-security-core, spring-web dependencies to mi…

eikenilsknopp:
…tigate CVE-2024-22257, CVE-2024-22259

### DIFF
--- a/sda-commons-dependencies/build.gradle
+++ b/sda-commons-dependencies/build.gradle
@@ -18,6 +18,11 @@ ext {
 }
 
 dependencies {
+  /* TODO: Next time this is updated, verify if the manual management of
+     org.springframework:spring-framework-bom:6.1.5 and org.springframework.security:spring-security-core:6.2.3
+     is still necessary.
+     https://docs.spring.io/spring-boot/docs/current/reference/html/dependency-versions.html
+   */
   api enforcedPlatform("org.springframework.boot:spring-boot-dependencies:${springBootVersion}"), {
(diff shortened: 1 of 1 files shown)


sda-commons-dependencies/build.gradle
```diff
@@ -32,13 +32,23 @@ dependencies {
...
 
   constraints {
     // overall testing dependencies
     api 'org.awaitility:awaitility:4.2.1'
 
+    // temporarily overwrite spring-boot dependencies to mitigate CVEs
+    api 'org.springframework.security:spring-security-core:6.2.3', {
+      because 'CVE-2024-22257'
+    }
+
```
JoergSiebahn, line 51: I assume, these managed dependency are usually managed [by Spring Boot Dependencies](https://github.com/SDA-SE/sda-spring-boot-commons/pull/655/files#diff-00048784048639f1f59b3085b1389df75870b87ca03587e5ce52ae6eeca1d49dR21)? If so, I would add a comment in that line to double check, so that we are aware when we see it in an update PR. Maybe mentioning the netty dependency there makes sense as well (if it is not already updated in Spring Boot).

eikenilsknopp: Yes, once spring-boot-dependencies contains the [necessary versions](https://docs.spring.io/spring-boot/docs/current/reference/html/dependency-versions.html) this can be removed again. I added a comment for this

JoergSiebahn (approved):

PR #723, opened 2024-05-30
feat: s3 health check

elbiocaetano:


sda-commons-starter-s3/src/main/java/org/sdase/commons/spring/boot/s3/config/S3HealthIndicator.java
```diff
@@ -0,0 +1,51 @@
...
+
+  @Override
+  protected void doHealthCheck(Health.Builder builder) {
+    try {
+      s3Client.headBucket(HeadBucketRequest.builder().bucket(bucketName).build());
+      builder.up().withDetails(Map.of("info", "S3 Bucket Available")).build();
+    } catch (Exception e) {
+      if (LOGGER.isErrorEnabled()) {
+        LOGGER.error(e.getMessage(), e);
+      }
```
JoergSiebahn, line 44: NIT: Imho this can be simplified, as `LOGGER.error` will also suppress the message if not enabled (and error is likely always enabled). I would also go for a custom message and attach the exception, e.g. like this:

```suggestion
      LOGGER.error("S3 health check failed to get info of bucket {}", bucketName, e);
```

JoergSiebahn: As the health check is propagating to the health endpoint anyways, I wonder if the log could also be a warning.

sda-commons-starter-s3/src/test/java/org/sdase/commons/spring/boot/s3/config/S3HealthIndicatorIntegrationTest.java
```diff
@@ -0,0 +1,90 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.test.annotation.DirtiesContext;
+import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.model.HeadBucketRequest;
+
+@SetSystemProperty(key = "management.health.s3.enabled", value = "true")
+@SpringBootTest(
+    classes = S3TestApp.class,
+    webEnvironment = WebEnvironment.RANDOM_PORT,
+    properties = {"auth.disable=true", "opa.disable=true", "management.server.port=8071"})
```
JoergSiebahn, line 33: NIT: As this is a SpringBootTest, isn't it possible to set it as property?

```suggestion
@SpringBootTest(
    classes = S3TestApp.class,
    webEnvironment = WebEnvironment.RANDOM_PORT,
    properties = {"auth.disable=true", "opa.disable=true", "management.server.port=8071", "management.health.s3.enabled=true"})
```

sda-commons-starter-s3/src/test/java/org/sdase/commons/spring/boot/s3/config/S3HealthIndicatorIntegrationTest.java
```diff
@@ -0,0 +1,90 @@
...
+@S3Test
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
```
JoergSiebahn, line 35: NIT: Can the dirty context be avoided if we don't use SetSystemProperty? Afaik, this slows down the test suite.

JoergSiebahn (approved):
Just some comments.

sda-commons-starter-s3/src/test/java/org/sdase/commons/spring/boot/s3/config/S3HealthIndicatorIntegrationTest.java
elbiocaetano: The problem is that I want to test when it is enabled and when it is not, so I need to run the server again for each test. In order to do that I either use this approach or I separate in two test classes, one with the health check enabled, and the other one disabled.

sda-commons-starter-s3/src/test/java/org/sdase/commons/spring/boot/s3/config/S3HealthIndicatorIntegrationTest.java
elbiocaetano: Unfortunately no, because otherrwise it would ovewrite the `@SetSystemProperty` in the `checkThatS3HealthCheckIsNotEnabled` test.

JoergSiebahn (approved):

PR #904, opened 2025-05-21
feat: add in junit pioneer for testing

codingsemen:


sda-commons-dependencies/build.gradle
```diff
@@ -37,6 +37,8 @@ dependencies {
     // overall testing dependencies
     api 'org.awaitility:awaitility:4.3.0'
 
+    api "org.junit-pioneer:junit-pioneer:2.3.0"
```
mibutec, line 40: In the services I added this dependency, I had to do an exclusion, otherwise there were conflicts with die JUnit version: https://github.com/SDA-SE/rise-kvs-delivery-service/blob/main/build.gradle#L92-L94

I would have expected the managed dependency also to include something like this. Could you check if the transitive dependencies from junit-pioneer are in sync with managed junit dependencies

Additional info: In services we do strict dependency resolving: https://github.com/SDA-SE/rise-kvs-delivery-service/blob/main/build.gradle#L49-L55

codingsemen:
I want to discuss 2 further things, what's the goto for declaring version in 

`ext {
  bouncycastleVersion = '1.80'
  logbackContribVersion = '0.1.5'
  springBootVersion = '3.4.5'
  springCloudVersion = '2024.0.1'
  scalaVersion = '2.13.16'
  swaggerCoreVersion = '2.2.32'
  victoolsVersion = '4.38.0'
}`

vs declaring it directly on the import

`api 'org.slf4j:slf4j-api:2.0.17'`

Kind of confusing why we have both. Would like to stick to one way or another. 
Also is there a convention on how to use String Literals in groovy? 
I see a lot of variation between using `'-text-'` and `"-text-"` - i am a huge fan of conventions and uniform style but it's obviously not a dealbreaker to have both.

mibutec (dismissed):

mibutec:
I don't know if there is a general rule on this, so I can only speak for myself:

I prefer the direct approach for better readability

`api 'org.slf4j:slf4j-api:2.0.17'`

But if there are multiple dependencies with the same version I switch to the ext approach to ensure both libs are in sync

```
ext {
  slf4jVersion = '2.0.17'
}

api "org.slf4j:slf4j-api:${slf4jVersion}"
api "org.slf4j:jcl-over-slf4j:${slf4jVersion}"
```

This might also explain why sometimes singlequotes and sometimes doublequotes are used. In singlequotes `${...}`won't get resolved

mibutec (approved):

PR #921, opened 2025-06-17
fix(deps): bump spring boot and spring cloud version up

YellowFlora:
bump spring boot to 3.5.0
bump spring cloud to 2025.0.0

BREAKING CHANGE: it can happen that a misalignment occurs between the version of the junit-platform-launcher provided by gradle and the version of the junit-jupiter-engine provided in the spring dependencies. In this case, gradle is not able to run the tests and throws an error. It's necessary to add testRuntimeOnly 'org.junit.platform:junit-platform-launcher' as a dependency in your build.gradle file then. That takes the version for the unit-platform-launcher from the dependency management.


YellowFlora:
> > It's necessary to add testRuntimeOnly 'org.junit.platform:junit-platform-launcher' as a dependency in your build.gradle file then.
> 
> Is it possible to ship this dependency in the required version with the module [sda-commons-web-testing](https://github.com/SDA-SE/sda-spring-boot-commons/tree/main/sda-commons-web-testing) to make it non-breaking? Or is `testRuntimeOnly` a scope, we can't provide with a dependency that is used as `testImplementation`? Would a transitive `runtimeOnly` become `testRuntimeOnly` when using our module as `testImplementation`?

based of this comment from spring boot I would say it's not recommend to do this https://github.com/spring-projects/spring-boot/issues/44613#issuecomment-2825053703 but if you say it's more important to not have a breaking change I can see what I can do

JoergSiebahn (approved):

PR #927, opened 2025-06-25
fix: force mongodb driver version to be 5.4.0

YellowFlora:


sda-commons-starter-mongodb/build.gradle
```diff
@@ -1,3 +1,9 @@
+configurations.configureEach {
+  resolutionStrategy {
+    force 'org.springframework.data:spring-data-mongodb:4.4.7' // the 4.5.x version uses the mongo driver v5.5.x but the driver needs to be 5.4.0 or lower, 5.5.0 is not compatible with document db 4
```
JoergSiebahn, line 3: Afaik this affects only this repo and not the consuming services. Isn't it possible to manage the version in the [dependency module](https://github.com/SDA-SE/sda-spring-boot-commons/blob/main/sda-commons-dependencies/build.gradle) - and probably just the MongoDB driver and not the whole Spring Data module?

YellowFlora: Yeah, had to experiment a bit because Spring is a bit strict regarding dependencies

PR #929, opened 2025-07-01
migrate to maven central publishing

JoergSiebahn:
I also cleaned up some warnings in the build.gradle. We should be ready for Gradle 9 now.

Seems like the link to the snapshots in the contributions file is currently not working (similar for dropwiazard-commons worked yesterday). It is the same pattern when you follow the links in the [repo root](https://central.sonatype.com/repository/maven-snapshots/). I was able to use the snapshots locally.

A [test deployment without release](https://github.com/SDA-SE/sda-spring-boot-commons/compare/86cc1390edfdbccb689bbf226a549f1b5899b04c..ce1141a3e0a8dffb5b6e69a54728fa03474f19a3) worked fine:

image


build.gradle
```diff
@@ -98,96 +97,52 @@ subprojects {
...
-
-            scm {
-              connection = 'scm:git:https://github.com/SDA-SE/sda-spring-boot-commons.git'
-              developerConnection = 'scm:git:https://github.com/SDA-SE/sda-spring-boot-commons.git'
-              url = 'https://github.com/SDA-SE/sda-spring-boot-commons'
-            }
+        licenses {
+          license {
+            name = 'MIT License'
+            url = 'https://raw.githubusercontent.com/SDA-SE/sda-spring-boot-commons/main/LICENSE'
```
YellowFlora, line 119: Not super important but if you wanna, could you update the year in the License File?

.github/mvn-verification/pom.xml
```diff
@@ -0,0 +1,91 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>org.sdase.test</groupId>
+  <artifactId>verify-maven-dependencies</artifactId>
+  <version>0.0.0-SNAPSHOT</version>
```
YellowFlora, line 9: The version here is irrelevant?

build.gradle
JoergSiebahn: Imho, we intentionally used the dash (2022-) with a blank end year, so that we never need to touch it again.

YellowFlora (approved):

PR #1085, opened 2026-02-25
IDEA feat(sda-commons-web-testing): add junit extensions to start MongoDB and Kafka in testcontainers


YellowFlora:


JoergSiebahn:
Nice idea. I wonder, if we need to provide code for that. Could we just manage the dependencies and add a unit test which we include in the docs as example? [Here](https://testcontainers.com/guides/testing-spring-boot-kafka-listener-using-testcontainers/#_write_test_for_kafka_listener) it looks pretty simple to create a test environment for Spring Boot with standard features from test containers.

YellowFlora:
> Nice idea. I wonder, if we need to provide code for that. Could we just manage the dependencies and add a unit test which we include in the docs as example? [Here](https://testcontainers.com/guides/testing-spring-boot-kafka-listener-using-testcontainers/#_write_test_for_kafka_listener) it looks pretty simple to create a test environment for Spring Boot with standard features from test containers.

If I still remember correctly I wrote this after optimizing the ram usage in some services during test running. The intention was to be able to only start one testcontainer for the whole test suite instead of an extra container for each test class.
Another point was the possibility to provide a standardized version for the images. But that could also be done simpler. 

It could be that it is a bit more complicated than necessary atm tho

